PR #1, opened 2015-02-19
Add .gitignore

wojtekmach:
Copied from: https://github.com/stackbuilders/octohat/blob/master/.gitignore


jsl:
Thanks @wojtekmach !

PR #2, opened 2015-04-23
Add GHC 7.10 support

sestrella:


sestrella:
@jsl could you take a look please

PR #3, opened 2015-06-18
Tidy up specs

fujimura:
Hi, this library looks great!

I'd like to propose some tidy ups in specs:

- Use `after_` for teardown
- Use `shouldReturn` to assert result of IO action


jsl:
Thanks @fujimura! Nice cleanup on the tests - I'll remember these tricks in the future. :)

PR #6, opened 2015-12-24
Add stack and change .gitignore

CristhianMotoche:
@jsl Could you take a look?
- First, I added stack configuration in order to upload it to Stackage.
- Second, I had to add the capability to set empty environmental variables using System.Posix.Env, because the setEnv function from System.Environment.Compat has an issue when try to set an environmental variable with an empty string (you can read it here http://hackage.haskell.org/package/base-compat-0.8.2/docs/System-Environment-Compat.html#v:setEnv). I wrote three unit test in order to test this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ cabal.config
 *.prof
 *.aux
 *.hp
+.stack-work/
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,0 +1,18 @@
+# For more information, see: https://github.com/commercialhaskell/stack/blob/release/doc/yaml_configuration.md
+
+# Specifies the GHC version and set of packages available (e.g., lts-3.5, nightly-2015-09-21, ghc-7.10.2)
+resolver: lts-3.19
+
+# Local packages, usually specified by relative directory name
+packages:
+- '.'
+
+# Packages to be pulled from upstream that are not in the resolver (e.g., acme-missiles-0.3)
+extra-deps:
+  - parseerror-eq-0.1.0.1
+
+# Override default flag values for local packages and extra-deps
+flags: {}
+
+# Extra package databases containing global packages
+extra-package-dbs: []


jsl:
Hi @CristhianMotoche thanks for the contribution. I have one question - don't we want this package to compile on Windows any more? It seems like bringing in the Unix dependency will break compatibility with that platform, doesn't it?

CristhianMotoche:
Yes, you're right. I think I could fix it with options that affect the C pre-processor. Could you tell me if that sounds like a good idea? I'll test it in a computer with Windows, and I'll tell you if that works. Thanks.

jsl:
What is the problem you're trying to solve? I may be able to help if I
better understand what you're trying to do.
>
>

-- 
Justin Leitgeb <http://justinleitgeb.com>
Twitter: @justinleitgeb <http://twitter.com/justinleitgeb>
Github: @jsl <https://github.com/jsl>

Co-Founder and CTO, Stack Builders Inc. <http://www.stackbuilders.com>

CristhianMotoche:
Oh, well the problem is with `setEnv` function which does `unsetEnv key` when the value for some key is empty (e.g. `setEnv "FOO" ""` won't set the environmental variable FOO, because the value is empty), that's because in Windows empty environmental variables don't exist.

CristhianMotoche:
Oh, the problem is with `setEnv` function which does `unsetEnv key` when the value for some key is empty (e.g. `setEnv "FOO" ""` won't set the environmental variable FOO, because the value is empty), that's because in Windows empty environmental variables don't exist.

jsl:
Right but to preserve compatibility with Windows, is that something we
can/want to change?

At any rate, seems like a separate issue from adding this package to stack.
What do you think about opening an issue on this repo to discuss, and
keeping this pr stack-specific so that we can merge it?

>

-- 
Justin Leitgeb <http://justinleitgeb.com>
Twitter: @justinleitgeb <http://twitter.com/justinleitgeb>
Github: @jsl <https://github.com/jsl>

Co-Founder and CTO, Stack Builders Inc. <http://www.stackbuilders.com>

CristhianMotoche:
Hey, I just tried it on Windows. It works, and I opened a new PR. May you can see it?

PR #7, opened 2016-01-02
Add options cpp

CristhianMotoche:


jsl:
I'll leave this open for now, but as we discussed, I think we want the behavior to be the same on Windows and *nix re: empty variable strings. I'll re-consider, though, if anyone else chimes in on this thread in support of this change.

habibalamin:
I'm going to chime in and say I also want this behaviour (i.e. empty variables to be able to be set). `APPLICATION_DB_PASSWORD` will often be blank on my development environment, that should be a basic feature. No offence to any Windows users, but I just want to say, I also resent the fact that the lowest common denominator (who I'm sure aren't the vast majority of users, but I couldn't tell you) is ruining this for the rest of us (though I know it's no one's fault in particular).

Besides, that's not entirely accurate; I think [this GHC bug ticket](https://ghc.haskell.org/trac/ghc/ticket/12494) will help explain things. In short, GHC was wrong on this one, because Windows _does_ in fact support empty environment variables (according to this ticket).

habibalamin:
`dotenv` comes with a `parseFile` function, so I'm making do with this for now:

```haskell
import System.Posix.Env (setEnv)

loadEnvFile = Dotenv.parseFile ".env" >>= mapM_ setEnvTuple
    where
        setEnvTuple = ($ True) . flip . uncurry $ setEnv
```

habibalamin:
I noted in [#44](https://github.com/stackbuilders/dotenv-hs/issues/44#issuecomment-299949327) that my workaround above won't work if you need to load your .env file with something like dotenv -f .env -- command (for example, with supporting scripts).

I realise the danger of adding complexity through feature toggles, but allowing the user to pick their poison seems like the right option for situations where neither option is entirely without issue. Maybe a `--unix`/[`--allow-empty-values`](https://github.com/stackbuilders/dotenv-hs/issues/44) flag.

If we do go on that train, maybe I'm being a scope creeper here, but a `.dotenvrc`/`.envrc` file with a list of default options would be the logical next step (common in the Ruby world where I come from). I wouldn't consider that a must-have (or even necessarily a should), though.

CristhianMotoche:
Closing this PR regardless https://github.com/stackbuilders/dotenv-hs/pull/51#issuecomment-309638056

PR #11, opened 2016-01-26
Upgrade to lts-5.0

CristhianMotoche:
I updated the stack.yml file to the lts-5.0. The library `parseerror-eq-0.1.0.1` is included in this lts, so I think we can remove it from the extra-dependency list in the stack.yml file.


jsl:
@CristhianMotoche thanks!

PR #12, opened 2016-02-23
Change .travis.yml to a container-based infrastructure

CristhianMotoche:


CristhianMotoche:
It's having the same problem that [inflections-hs](https://github.com/stackbuilders/inflections-hs/pull/22#issuecomment-171077501) with hspec-discover.

PR #13, opened 2016-02-23
Change travis

CristhianMotoche:

I also had to change the test, so that they can run in any GHC version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,85 @@
-env:
-  - GHC_VERSION=7.4.2  CABAL_VERSION=1.18
-  - GHC_VERSION=7.6.3  CABAL_VERSION=1.20
-  - GHC_VERSION=7.8.4  CABAL_VERSION=1.20
-  - GHC_VERSION=7.10.1 CABAL_VERSION=1.22
+# This file has been generated -- see https://github.com/hvr/multi-ghc-travis
+language: c
+sudo: false
+
+cache:
+  directories:
+    - $HOME/.cabsnap
+    - $HOME/.cabal/packages
+
+before_cache:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,21 +1,85 @@
...
+      addons: {apt: {packages: [cabal-install-1.16,ghc-7.4.2], sources: [hvr-ghc]}}
+    - env: CABALVER=1.16 GHCVER=7.6.3
+      compiler: ": #GHC 7.6.3"
+      addons: {apt: {packages: [cabal-install-1.16,ghc-7.6.3], sources: [hvr-ghc]}}
+    - env: CABALVER=1.18 GHCVER=7.8.4
+      compiler: ": #GHC 7.8.4"
+      addons: {apt: {packages: [cabal-install-1.18,ghc-7.8.4], sources: [hvr-ghc]}}
+    - env: CABALVER=1.22 GHCVER=7.10.1
+      compiler: ": #GHC 7.10.1"
+      addons: {apt: {packages: [cabal-install-1.22,ghc-7.10.1], sources: [hvr-ghc]}}
```
jpvillaisaza: @CristhianMotoche We can change this to 7.10.2.

Are all four of these taking too long? I guess we don't have to worry about 7.4.2.

In line 73, we can remove the `--enable-benchmarks` I think (not sure if it matters).

Other than these comments, this looks good.

CristhianMotoche: Ok, I'll make those changes. Thanks.

CristhianMotoche: I'll close this PR and I'll open a new one, now that I have permissions.

CristhianMotoche:
Redirect  to #14

PR #14, opened 2016-02-23
Change travis

CristhianMotoche:

I also had to change the test, so that they can run in any GHC version.


CristhianMotoche:
I'm going to close this PR if we merge #16

PR #18, opened 2016-03-20
Add Text version of parsefile along with String version

fujimura:
Functions in `System.Environment`, `System.Process` etc. returns or takes environment variables as `String`, not `Text`.
In many cases users will merge them with ones from dotenv, so it'll be better to leave interfaces in plain String for their convenience.


jsl:
@fujimura ok - I can see how it makes sense to leave this as String. Thank you for also adding the file for the Text interface.

jsl:
Pushed version 0.3.0.0 to hackage with these changes. Thanks again!

fujimura:
Thanks a lot!

Issue #19, opened 2016-03-20
Remove extra dependencies from stack.yaml


jpvillaisaza:
There are three extra dependencies in `stack.yaml`: `fail-4.9.0.0`, `megaparsec-4.4.0`, and `semigroups-0.18.1`. We can remove `fail` and `semigroups` from this file. I'm not sure about `megaparsec` -- it depends on whether the library builds with `megaparsec-4.3.0` or not.

jsl:
I only added those reluctantly after not being able to work out some build failures last night. If you try removing them, and push a branch I think you'll see what I mean with the Travis results. If you can find a way around them, or if those failures don't matter I'd love to remove them.

jsl:
Specifically, it won't build at least with lts-3 unless you add some dependencies. You can do this locally by building with `stack build --resolver lts-3`. Maybe this doesn't matter, since I don't think dotenv is part of LTS 3 for Stackage?

```
$ stack solver --resolver lts-3
Selected resolver: lts-3.22
Using configuration file: stack.yaml
Using cabal packages:
- dotenv.cabal

Using resolver: lts-3.22
Using compiler: ghc-7.10.2
Asking cabal to calculate a build plan...
Trying with packages from lts-3.22 as hard constraints...
Successfully determined a build plan with 2 external dependencies.

The following changes will be made to stack.yaml:
* Resolver is lts-3.22
* Dependencies to be added
    extra-deps:
    - megaparsec-4.3.0
    - text-1.2.2.0

To automatically update stack.yaml, rerun with '--update-config'
```

jpvillaisaza:
@jsl I was hoping Stack allowed to add extra dependencies using a command line option, but it seems like the `extra-deps` section in the `stack.yaml` file is the only option.

This is failing for only two out of ten build jobs and the solution is a workaround for a (historical) limitation in LTS Haskell 3. I think the fact that Cabal with GHC 7.6.3, 7.8.4, and 7.10.3 is able to solve dependencies for both Linux and OS X is enough to say that these failures don't matter -- we know they work if you add the extra dependencies.

jsl:
Closing, thanks for taking care of this in #20, @jpvillaisaza!

PR #20, opened 2016-03-20
Remove extra dependencies from stack.yaml

jpvillaisaza:
@jsl This is what I had in mind regarding issue #19 (and https://github.com/stackbuilders/dotenv-hs/issues/19#issuecomment-198982228 in particular).

(Not a big deal, but it takes around eight minutes less to finish all build jobs.)


jsl:
Looks good to me, thanks! I'll merge this now, and push a new release to hackage later this evening.

PR #28, opened 2016-08-16
Allow optparse-applicative 0.13

jpvillaisaza:
See fpco/stackage#1784.


jpvillaisaza:
@jsl This is enough for using the latest optparse-applicative. The only change is importing `<>` from `Data.Monoid`, since optparse-applicative is no longer re-exporting it.

jsl:
Perfect, LGTM! Thanks @jpvillaisaza.

Issue #29, opened 2016-10-10
Expand existing variables


jpvillaisaza:
I would like to use a .env file like the following:

```
FOO=${HOME}/.foo
```

After loading the file, `FOO` should be `/home/jpvi/.foo` or something like that.

@jsl Let me know if this makes sense and I can take a look.

jsl:
@jpvillaisaza I think that this is a good initiative. Thanks for opening this issue. We should think about how other libraries implement this sort of feature, and think critically about the best way to proceed.

I noticed that:

1. Dotenv for Node.js has variable expansion [as a separate module](https://github.com/motdotla/dotenv-expand).
2. Ruby dotenv does [tilde expansion](https://github.com/bkeepers/dotenv/blob/master/spec/dotenv_spec.rb#L19:L29).
3. Python has notes about possible [POSIX-compliant variable expansion](https://github.com/theskumar/python-dotenv/pull/30#issuecomment-244036604).

I agree with the general idea of adding functionality to the dotenv templates. However, we may just want to proceed slowly, and based on standards wherever possible so that we don't introduce something that we regret later. We may also want to separate the notion of path expansion from full POSIX-compliant variable expansion and consider those two separately.

What do you think?

jpvillaisaza:
@jsl Yes, sounds good. I saw the issue for the Python one before opening this issue. I don't have experience with this library (use or development), so I can start taking a look and we can add notes here to think of the best way.

habibalamin:
I ran a test using the `dotenv`gem, and I think their implementation is sensible. Given how many different implementations of `dotenv` there are in different languages, I think a standard may be useful to codify.

`.env` file:

```sh
START='this is a test' # test with spaces
REFERENCE_TO_START=$DOTENV_A

UNQUOTED_REFERENCE=$HOME
UNQUOTED_REFERENCE_WITH_EXTRA=$HOME_abc
UNQUOTED_REFERENCE_WITH_BRACES=${HOME}
UNQUOTED_REFERENCE_WITH_BRACES_AND_EXTRA=${HOME}_abc

SINGLE_QUOTED_REFERENCE='$HOME'
SINGLE_QUOTED_REFERENCE_WITH_EXTRA='$HOME_abc'
SINGLE_QUOTED_REFERENCE_WITH_BRACES='${HOME}'
SINGLE_QUOTED_REFERENCE_WITH_BRACES_AND_EXTRA='${HOME}_abc'

[...]

PR #42, opened 2017-02-23
Expand existing variables

habibalamin:
This is my crack at #29. It's not in a working state, as I have zero experience with parsing before this (and I'm still quite new to Haskell as well). Currently, I get an error like so when running `parseFile`:

```
*** Exception: Text.Megaparsec.Prim.many: combinator 'many' is applied to a parser that may consume no input.
CallStack (from HasCallStack):
  error, called at ./Text/Megaparsec/Prim.hs:376:11 in megaparsec-5.0.1-Kbr1rj2aE5WBrRq6kJu93J:Text.Megaparsec.Prim
```


jsl:
Hi @habibalamin, thanks very much for your contribution! @mrkkrp is more experienced with the parsing code in dotenv than I am at this point. Perhaps he has some feedback?

src/Configuration/Dotenv/Parse.hs
```diff
@@ -16,43 +16,65 @@
...
 
 -- | Value: quoted or unquoted.
-value :: Parser String
+value :: Parser VValue
 value = (quotedValue <|> unquotedValue) <?> "variable value"
   where
-    quotedValue   = quotedWith '\'' <|> quotedWith '\"'
-    unquotedValue = many (noneOf "\'\" \t\n\r")
+    quotedValue   = quotedWith SingleQuote <|> quotedWith DoubleQuote
+    unquotedValue = Unquoted <$> (many $ fragment "\'\" \t\n\r")
```
habibalamin: I think the issue is with the use of `many` with `fragment` (also in `quotedWith`, which does the same thing).

mrkkrp: This happens because you should never nest one `many` into another one. Since `many` parses *zero* or more occurences, it succeeds even when its argument parser doesn't match at at all. And if something succeds without consuming input/advancing parser in `many`, it'll run it indefinitely. First `many` is where you say, and the second (inner) `many` is in `literalValueFragment`. I propose the following technique:

* Use `some` intead of many, so fragments can't be empty.
* Empty values just have zero fragments: `[]`.

There are other minor issues, but they are not important at this point, we will talk about them when the PR is more complete. BTW, good work for a person new to Haskell!

habibalamin:
I have started working on this again, but I'm really kind of busy with family stuff and another, higher priority project that I haven't found the time to finish this off (as you said, there are a few problems with the PR as it is).

If anyone wants to pick this up, they're welcome to, but just say so here so I don't keep working on it on my downtime.

habibalamin:
Took a couple hours, but I got it (after I realised I could play around with the miniparsers in GHCi). @mrkkrp anything you want me to do before this can be incorporated?

habibalamin:
You can test with this:

```
# .env
ABC="$HOME"
DEF="${HOME}"
GHI='$HOME'
JKL=HOME
MNO=$HOME
PQR=''
STU=""
VWX=
YZ='still'
```

```bash
# test.sh

# Run with below lines commented and uncommented, just to prove that the shell works the same way.
# ABC="$HOME"
# DEF="${HOME}"
# GHI='$HOME'
# JKL=HOME
# MNO=$HOME
# PQR=''
# STU=""
# VWX=
# YZ='still'

echo $ABC
echo $DEF
echo $GHI
echo $JKL
echo $MNO
echo $PQR
echo $STU
echo $VWX
echo $YZ
```

Test with `.env` by:

```
stack exec -- dotenv -f .env ./test.sh
```

Test the shell's (bash's?) parser by uncommenting the variable setters and running (with `set -u`, if you like, to make sure the shell does bind the variables):

```
./test.sh
```

habibalamin:
Fixing tests now.

src/Configuration/Dotenv/ParsedVariable.hs
```diff
@@ -0,0 +1,55 @@
...
+                                            interpolateParsedVariables) where
+
+import Control.Monad (foldM)
+#if MIN_VERSION_optparse_applicative(0,13,0)
+import Data.Functor ((<$>))
+#endif
+import Control.Applicative ((<|>))
+import System.Environment (lookupEnv)
+
+data ParsedVariable
```
mrkkrp: I would still like to see a better naming scheme here. Let's at least rename the `V` prefix to `Var` or `Variable`.

spec/Configuration/DotenvSpec.hs
```diff
@@ -8,6 +8,8 @@ import Test.Hspec
 
 import System.Environment (lookupEnv)
 import Control.Monad (liftM)
+import Data.Maybe (fromMaybe)
+import Data.Functor ((<$>))
```
juanpaucar: This could be conditionally imported with [this](https://github.com/stackbuilders/dotenv-hs/blob/master/spec/Configuration/DotenvSpec.hs#L13)

src/Configuration/Dotenv/Parse.hs
```diff
@@ -16,43 +16,60 @@
...
 import Text.Megaparsec.String (Parser)
 import qualified Text.Megaparsec.Lexer as L
 
+data QuoteType = SingleQuote | DoubleQuote
+
 -- | Returns a parser for a Dotenv configuration file. Accepts key and value
 -- arguments separated by @=@. Comments in all positions are handled
 -- appropriately.
-configParser :: Parser [(String, String)]
+configParser :: Parser [ParsedVariable]
```
juanpaucar: Should we add to the documentation above this  function that now it also accepts quoted and interpolated values?

src/Configuration/Dotenv/ParsedVariable.hs
```diff
@@ -0,0 +1,55 @@
...
+  | SingleQuoted VContents
+  | DoubleQuoted VContents deriving (Show, Eq)
+
+type VContents = [VFragment]
+
+data VFragment
+  = VInterpolation String
+  | VLiteral String deriving (Show, Eq)
+
+interpolateParsedVariables :: [ParsedVariable] -> IO [(String, String)]
```
juanpaucar: Do we need to have `IO` wrapping all these new functions? It seems unnecessary. They could pure functions
- interpolateParsedVariables
- addInterpolated
- interpolate
- interpolateContents
- interpolateFragment

\cc @mrkkrp

src/Configuration/Dotenv/ParsedVariable.hs
```diff
@@ -0,0 +1,55 @@
...
+interpolateParsedVariables = fmap reverse . foldM addInterpolated []
+
+addInterpolated :: [(String, String)] -> ParsedVariable -> IO [(String, String)]
+addInterpolated previous (ParsedVariable name value) = (: previous) <$> ((,) name <$> interpolate previous value)
+
+interpolate :: [(String, String)] -> VValue -> IO String
+interpolate _        (SingleQuoted contents) = return $ joinContents contents
+interpolate previous (DoubleQuoted contents) = interpolateContents previous contents
```
juanpaucar:
```haskell
interpolate previous (DoubleQuoted contents) = interpolateContents previous contents
interpolate previous (Unquoted     contents) = interpolateContents previous contents
```
could be?
```haskell
interpolate previous (_ contents) = interpolateContents previous contents
```

mrkkrp: You can't pattern-match like this I think :-D

juanpaucar: my bad

src/Configuration/Dotenv/Parse.hs
habibalamin: I don't think this is the right place for that piece of information.

src/Configuration/Dotenv/ParsedVariable.hs
habibalamin: I'll take another crack at it.

src/Configuration/Dotenv/Parse.hs
juanpaucar: Sorry, What I meant is to update the commentary for this function to include what is now being added

src/Configuration/Dotenv/ParsedVariable.hs
habibalamin: I typed up a relatively lengthy resposne to this, did you get it? It seems to have disappeared.

src/Configuration/Dotenv/Parse.hs
habibalamin: Sorry, I don't follow; what's the difference between what you said before and what you're saying now?

src/Configuration/Dotenv/ParsedVariable.hs
habibalamin: I'll just type up the response as best as I can remember it.

As I said, I'm quite new to Haskell, so I'm not exactly well-versed in keeping functions as pure as possible. I took a quick look again to reacquaint myself with the code I wrote here 2 months ago, and the IO does seem to be necessary in all the functions (and the necessity derives from the fact that we access the environment).

I'd love some advice on maybe what abstraction would work well to achieve the effect I'm trying to achieve here, but `foldM` was the simplest way I could see to write the `interpolateParsedVariables` function.

I can't really be bothered to explain why the rest of them need IO as well one by one all over again, but if you have any ideas, I'm game to hear them.

mrkkrp: OK, I've looked at the code and it seems that:

1. The whole `Configuration.Dotenv.ParsedVariable` module is an internal helper, so it's sort of OK to have `IO` there.

2. The only place this `interpolateParsedVariables` function is used is the `parseFile`, which is already in `MonadIO`.

So as long as `interpolateParsedVariables` only lookups things and does not modify environment, I'm OK with the current signatures. We of course could grab all the environment as a `Map String String` and then have a reader monad instead of `IO`, but for this use case, it looks like an unjustified and slower solution.

src/Configuration/Dotenv/Parse.hs
habibalamin: Ha. I understand now. That's a Haddock comment.

Issue #41, opened 2016-11-16
Make a "safe" wrapper for loading variables


jsl:
Following the example of dotenv-safe, we can be more helpful when loading environment data. We may be able to define types of variables to import, to specify variables that need to be specified, whether or not they need to have non-null contents, and perhaps other basic types.

One possible implementation that I discussed with @sestrella was that we may be able to allow the user to specify a template, perhaps in dotEnv format. The left hand side would specify variable names that could occur, and the right hand side may point to a type that could contain a parser defined in Megaparsec that is defined internally in dotenv. We should be able to specify whether or not a variable should be mentioned at all in the input file, and if it is mentioned, can it be null? Otherwise, should it be an integer, alphanumeric string, etc.

jpvillaisaza:
The idea is https://www.npmjs.com/package/dotenv-safe, right?

@well1791 I think you and @sestrella have an initial idea ready. Would it make sense to add a comment here to discuss it and then work on it?

well1791:
@jpvillaisaza Good idea.

@sestrella suggests me that we can define something like a _schema_ file to compare with the actual _.dotenv_ file e.g.

_.dotenv.schema_
```json
{
  "DB_HOST": {
    "required": true,
    "type": "string",
    "allowEmpty": false
  }
}
```

_.dotenv_
```
DB_HOST=localhost
```

So the operation that performs the checking will expect the path of those two files.

@jpvillaisaza what do you think?
\cc @sestrella @jsl

jpvillaisaza:
Does it make sense to add command line options as well (perhaps as part of a separate issue/pull request)? I think it could make sense for an option like `--allow-empty-values` (which would affect all variables specified in the sample file).

Other than that, I guess the default value for an environment variable will be to require it (like dotenv-safe does). Thus, I'm not sure including both `required` and `allowEmpty` values is necessary -- just `allowEmpty` (or `allowEmptyValues`) should be enough (and the default value is false, which is like setting a required value to true).

Also, will this take place over the current executable or will it be two executables, one for an unsafe dotenv and another one for a safe version?

well1791:
@jpvillaisaza I think it will be ok in the current executable, we can find if exists that `.dotenv.schema.json` file, and do the checking if we find it.

sestrella:
[...]

PR #43, opened 2017-03-07
Add dotenv-safe functionality

acamino:
This PR would solve issue #41


app/Main.hs
```diff
@@ -8,19 +9,32 @@ import Data.Monoid ((<>))
...
-  { files    :: [String]
-  , program  :: String
-  , overload :: Bool
+  { program           :: String -- ^ Program to run with the load env variables
+  , dotenvFile        :: String -- ^ Path for the .env file
+  , dotenvExampleFile :: String -- ^ Path for the .env.example file
+  , override          :: Bool   -- ^ Override current environment variables
   } deriving (Show)
 
+buildConfig :: FilePath -> FilePath -> Config
```
mrkkrp: Given that we do not export `Options`, I think this `buildConfig` function in redundant, we can just do everything in place in `dotEnv`. (I there is a different `buildConfig` in tests, that one is probably handy and can stay.)

One thing I'd like to see in SB Haskell code is lesser number of functions. When code is fragmented, it's more difficult to read. In this particular case, the `dotEnv` function is not necessary. When I read `main`, I just want to be able to read it all in one place. So why not write it this way:

```haskell
main :: IO ()
main = do
  Options {..} <- execParser opts -- opts could be a stand-alone top-level definition
  current <- getCurrentDirectory
  -- the rest goes here
```

app/Main.hs
```diff
@@ -30,21 +44,34 @@ main = execParser opts >>= dotEnv
...
-                  <> metavar "FILE"
-                  <> help "File to read for options" ))
+config =
+  Options
+     <$> argument str (metavar "PROGRAM")
+
+     <*> strOption
+          ( long "env"
+          <> value ".env"
+          <> metavar "\"DOTENV FILE\""
```
mrkkrp: We could use `showDefault` here. Also: I don't understand why you need to quote manually contents of `metavar`. Maybe just `"DOTENV_FILE"` will do?

app/Main.hs
```diff
@@ -30,21 +44,34 @@ main = execParser opts >>= dotEnv
...
+          <> help "File with the env variables" )
 
-     <*> argument str (metavar "PROGRAM")
+     <*> strOption
+          ( long "env-example"
+          <> value ".env.example"
+          <> metavar "\"DOTENV EXAMPLE\" FILE"
```
mrkkrp: The same here, `showDefault` and not sure why you need quotes here.

app/Main.hs
```diff
@@ -30,21 +44,34 @@ main = execParser opts >>= dotEnv
...
+          <> help "File with all the necesary env variables" )
 
-     <*> switch ( long "overload"
-                  <> short 'o'
-                  <> help "Specify this flag to override existing variables" )
+     <*> switch
+           ( long "overload"
```
mrkkrp: It would be good to be consistent in wording: is it "overload" or "override"? If it's overload, then we could edit name of record selector in `Options` to match.

app/Main.hs
```diff
@@ -30,21 +44,34 @@ main = execParser opts >>= dotEnv
...
+           <> short 'o'
+           <> help "Specify this flag to override existing variables" )
```
mrkkrp: I think it's better to drop the `"Specify this flag to"` part.

app/Main.hs
```diff
@@ -30,21 +44,34 @@ main = execParser opts >>= dotEnv
...
 
 dotEnv :: MonadIO m => Options -> m ()
-dotEnv opts = liftIO $ do
-  mapM_ (loadFile (overload opts)) (files opts)
-  code <- system (program opts)
+dotEnv Options{..} = liftIO $ do
+  current <- getCurrentDirectory
+
+  let defaultConfig = buildConfig (current </> dotenvFile) (current </> dotenvExampleFile)
```
mrkkrp: What's the point of prepending `current` to all the paths? If given path is relative, it will be relative to current directory automatically. It should work without `getCurrentDirectory` and `current`, unless I'm missing something.

dotenv.cabal
```diff
@@ -38,7 +38,7 @@ maintainer:          hackage@stackbuilders.com
 copyright:           2015-2016 Stack Builders Inc.
 category:            Configuration
 build-type:          Simple
-extra-source-files:    spec/fixtures/.dotenv
+extra-source-files:    test/fixtures/.dotenv
```
mrkkrp: Why renaming? It would help to have as few unnecessary changes as possible.

dotenv.cabal
```diff
@@ -89,14 +94,12 @@ library
 
 test-suite dotenv-test
   type: exitcode-stdio-1.0
-  hs-source-dirs: spec, src
+  hs-source-dirs: test
```
mrkkrp: This is good, `src` directory has no business here.

src/Configuration/Dotenv.hs
```diff
@@ -1,86 +1,66 @@
...
 --
--- This module contains common functions to load and read dotenv files.
+-- Provides the ability to load a dotenv file with an specific configuration.
 
-{-# LANGUAGE CPP #-}
+{-# LANGUAGE CPP             #-}
+{-# LANGUAGE RecordWildCards #-}
 
 module Configuration.Dotenv
-  ( load
```
mrkkrp: I'm concerned with backward compatibility here. I think we should not break it like this. I know a number of projects now that rely on the functions you deleted. And for some functions you do not provide a sensible and handy replacement. I remember adding `onMissingFile` and it's very useful.

More generally, I do think we should add the new functionality so it coexists with existing functionality, not replace it. I'd like my app to error early if a variable is not present in some cases, not use some default value. In some cases there is no good default.

src/Configuration/Dotenv/Parser.hs
```diff
@@ -14,22 +14,24 @@
 
 {-# LANGUAGE CPP #-}
 
-module Configuration.Dotenv.Parse (configParser) where
```
mrkkrp: Is this a whim? We should not rename modules of published packages like this. This even doesn't help anything and breaks compatibility, this should be renamed back.

src/Configuration/Dotenv/File.hs
```diff
@@ -0,0 +1,38 @@
...
+import Text.Megaparsec
+
+import Configuration.Dotenv.Parser
+import Configuration.Dotenv.Types
+
+-- | @parseMaybeFile@ parses a @.env@ file.
+-- If the file does not exist then it returns @Nothing@.
+parseMaybeFile :: MonadIO m => FilePath -> m (Maybe [Variable])
+parseMaybeFile path = liftIO $ do
+  exist <- doesFileExist path
```
mrkkrp: The `onMissingFile` approach is better because if it's not a subject to race conditions. If you test just before opening a file, you still may find that file missing a microsecond after. You should try to open it once and deal with exceptions, like it's done in `onMissingFile`. As I said I strongly recommend to put those functions you deleted back and preserve maximum backward-compatibility.

src/Configuration/Dotenv/Types.hs
```diff
@@ -0,0 +1,22 @@
...
+-- Maintainer  :  Stack Builders <hackage@stackbuilders.com>
+-- Stability   :  experimental
+-- Portability :  portable
+--
+-- Provides the types with extra options for loading a dotenv file.
+
+module Configuration.Dotenv.Types where
+
+-- | Alias for environment variables (NAME, VALUE).
+type Variable = (String, String)
```
mrkkrp: I'd prefer `(String, String)`. Reason: when I see `Variable` I think: "Hmm, and what *actually* this is?". I need to look up. With `(String, String)` I see what is this and how it's organized at a glance.

app/Main.hs
acamino: Thanks for the suggestion. You're right, `"DOTENV_FILE"` is more than enough.

dotenv.cabal
acamino: +1

src/Configuration/Dotenv/Parser.hs
acamino: Thanks for pointing out to the possible issues that this change might introduce. Agreed, I'll rename back this one.

CristhianMotoche:
Hey @mrkkrp, this is ready for another review.
Thanks!

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## Dotenv 0.4.0.0
+
+* Add `parseMaybeFile` function for parsing dotenv files.
+
+* Changed `loadFile` function to get `Config` with the paths for the .env file
+  and the .env.example file.
+
+* Add the `Config` type with options to override env variables, and setting the
```
sestrella: @acamino :+1:

src/Configuration/Dotenv.hs
```diff
@@ -1,78 +1,76 @@
...
-parseFile ::
-  MonadIO m =>
-  FilePath -- ^ A file containing options to read
-  -> m [(String, String)] -- ^ Variables contained in the file
-parseFile f = do
-  contents <- liftIO $ readFile f
+setupEnvVars :: MonadIO m => Bool -> [String] -> Maybe [NameValuePair] -> m [NameValuePair]
+setupEnvVars override keys maybeVars =
+  case maybeVars of
+    Nothing   -> mapM getEnvVariable keys
```
sestrella: @acamino you don't need to pattern match on `Nothing` what about:

```Haskell
case maybeVars of
  Just vars -> setVariables override keys vars
  _ -> mapM getEnvVariable keys
```

src/Configuration/Dotenv.hs
```diff
@@ -1,78 +1,76 @@
...
-  if override then
-    liftIO $ setEnv key value
+setVariables :: MonadIO m => Bool -> [String] -> [NameValuePair] -> m [NameValuePair]
+setVariables override keys vars =
+  mapM (getVariable vars) keys >>= mapM (setVariable override)
+
+getVariable :: MonadIO m => [NameValuePair] -> String -> m NameValuePair
+getVariable vars key =
+  case lookup key vars of
+    Nothing  -> getEnvVariable key
```
sestrella: @acamino same here

src/Configuration/Dotenv.hs
```diff
@@ -1,78 +1,76 @@
...
+setVariable override var@(key, value) =
+  if override then
+    liftIO $ setEnv key value >> return var
   else do
     res <- liftIO $ lookupEnv key
-
     case res of
-      Nothing -> liftIO $ setEnv key value
-      Just _  -> return ()
+      Nothing   -> liftIO $ setEnv key value >> return var
```
sestrella: @acamino same here

stack.yaml
```diff
@@ -1,3 +1 @@
-resolver: lts-7.4
-packages:
-- '.'
+resolver: lts-8.0
```
sestrella: @acamino what about using latest `lts` version `8.4`?

src/Configuration/Dotenv.hs
acamino: You're right - I dodged it.

stack.yaml
acamino: +1

CristhianMotoche:
Hey @sestrella, This is ready for another review.
Thanks!

src/Configuration/Dotenv.hs
```diff
@@ -1,78 +1,76 @@
...
--- override existing variables with values from Dotenv files.
-loadFile ::
-  MonadIO m =>
-  Bool        -- ^ Override existing settings?
-  -> FilePath -- ^ A file containing options to load into the environment
-  -> m ()
-loadFile override f = load override =<< parseFile f
+-- | @loadFile@ parses the environment variables defined in the @dotenv example@
+-- file and checks if they are defined in the @dotenv@ file or in the environment.
+-- It also allows to @override@ the environment variables defined in the environment
```
mrkkrp: Why do you use `@override@` syntax here? What is `override`?

src/Configuration/Dotenv/Types.hs
```diff
@@ -0,0 +1,22 @@
...
+-- Maintainer  :  Stack Builders <hackage@stackbuilders.com>
+-- Stability   :  experimental
+-- Portability :  portable
+--
+-- Provides the types with extra options for loading a dotenv file.
+
+module Configuration.Dotenv.Types where
+
+-- | Alias for environment variables (NAME, VALUE).
+type NameValuePair = (String, String)
```
mrkkrp: I'd still propose to use bare `(String, String)` for reasons I mentioned earlier.

CristhianMotoche: I think `NameValuePair` add semantical meaning, a little bit more than `(String, String)`. We can discuss this before merge.
Thanks @mrkkrp

src/Configuration/Dotenv.hs
CristhianMotoche: I'll check a little bit of Haskell code documentation and fix that.
Thanks!

src/Configuration/Dotenv/Types.hs
mrkkrp: Do as you wish, it's just `(String, String)` is already an obivous pair and it's absolutely clear what it represents and how; keeping it as is saves users a lookup. This is my opinion is the synonym doesn't help anything.

CristhianMotoche: Ok, then, I think it will be fine to use just `(String, String)`

CristhianMotoche:
@acamino @mrkkrp @sestrella Do you have any other comments on this PR? Otherwise, we can merge it. 😄

CristhianMotoche:
@acamino I think we should squash some commits and then merge this PR. I've already rebased it with master.

CristhianMotoche:
The squash is done. It is ready to merge and bump in Hackage. 😄

CristhianMotoche:
I've noticed that we've lost a functionallity with this new feature: we won't be able to set the env vars from multiple `.env`

CristhianMotoche:
@sestrella @mrkkrp @acamino This is ready for a last review. 😄

juanpaucar:
@CristhianMotoche @acamino Please check the merging conflicts

CristhianMotoche:
@juanpaucar sure I'll do it. I hope you can take a look at this and give us feedback to merge it or fix it if necessary. Thanks.

CristhianMotoche:
@juanpaucar Could you take a look please?

app/Main.hs
```diff
@@ -8,20 +9,33 @@ import Data.Monoid ((<>))
...
+import Configuration.Dotenv.Types
+
 data Options = Options
-  { files    :: [String]
-  , overload :: Bool
-  , program  :: String
-  , args     :: [String]
+  { program           :: String -- ^ Program to run with the load env variables
+  , args              :: [String] -- ^ Args for the program that is going to be executed
+  , dotenvFile        :: String -- ^ Path for the .env file
```
habibalamin: I know this has been approved, but I just wanted to ask, is this change necessary? What's the purpose of it?

habibalamin: Never mind.

This PR's initial comment doesn't link to any issues or contain an explanation, so I was confused on its purpose at first. I just realised it might be referring to another package named `dotenv-safe` (looks like a Node package).

Could you not still have multiple environment files and one example file?

CristhianMotoche: @habibalamin It is related to #41. However, we've been discussing about this change and we want to add a flag (`--safe`) to behave this way. That is why this PR is on hold for now.

CristhianMotoche:
Since we conserved `onMissingFile` I think we can omit `parseMaybeFile`
```haskell
listToMaybe <$> onMissingFile (parseFile "does/not/exist.txt") (return [])
>>> Nothing
```
Also, I'm not sure why we want to `return Nothing` when the file is missing, it should failed!

juanpaucar:
@CristhianMotoche Are we still working on this?

CristhianMotoche:
@juanpaucar I'm not sure what will be the future of this PR. As we've discussed this feature only brings the capability to be sure that the mandatory env vars (defined in `.env.example`) are set. However, we also want to add some kind of "type safe env vars". These things stopped me to continue.

juanpaucar:
Let's stick to the discussion on the open issue. We could tackle type safe environment vars later and we can have a broader discussion only focused on that

CristhianMotoche:
🆒

CristhianMotoche:

juanpaucar:
@CristhianMotoche could you update the readme file with the usage examples, please

src/Configuration/Dotenv/Types.hs
```diff
@@ -0,0 +1,30 @@
...
+-- License     :  MIT
+--
+-- Maintainer  :  Stack Builders <hackage@stackbuilders.com>
+-- Stability   :  experimental
+-- Portability :  portable
+--
+-- Provides the types with extra options for loading a dotenv file.
+-}
+
+module Configuration.Dotenv.Types where
```
juanpaucar: we should be explicit about what we export on form this module

src/Configuration/Dotenv/File.hs
```diff
@@ -0,0 +1,27 @@
...
+-- License     :  MIT
+--
+-- Maintainer  :  Stack Builders <hackage@stackbuilders.com>
+-- Stability   :  experimental
+-- Portability :  portable
+--
+-- This module provides parsers for dotenv files.
+-}
+
+module Configuration.Dotenv.File where
```
juanpaucar: can we be specific about what we export from tis module, please?

CristhianMotoche:
Thanks @juanpaucar. I'll do a little refactor today in the evening and update the README.

CristhianMotoche:
@juanpaucar Ready for a new review!

src/Configuration/Dotenv.hs
jsl: @CristhianMotoche it seems we're still breaking backward compatibility here without providing a replacement. Why are we doing this? Are we sure no one is using this functionality?

CristhianMotoche: @jsl I'll bring that functionally back.

juanpaucar:
This PR has already too many comments. The functionality has been added in another PR and the version was increased as it is a breaking change for the library.

I'm closing this in favor of #57

PR #50, opened 2017-05-12
Allow blank variables on POSIX and Windows

### LINKED ISSUE #48: Blank environment variables aren't loaded (even in a POSIX environment)
I think being able to have blank environment variables is useful, and I kind of resent the fact that Windows is acting as a lowest common denominator here (no offence to the team here, I can understand why you made the decision).

Some conversation occurred on #7 and #44.

---

For those coming to this issue after a lot of conversation, [this bug is being tracked upstream](https://ghc.haskell.org/trac/ghc/ticket/12494), as it's a bug with GHC itself. I'll update this comment to keep track of the status of the bug.

habibalamin:
This is a work in progress to fix #48. I did get it working for POSIX (on macOS), and I have now got it compiling for Windows, but when calling `setEnv`, it doesn't seem to have any effect on the environment.

I had the same issue with the POSIX portion when I was working on that, so I don't think it's anything to worry about. Interestingly enough, doing a `lookupEnv`, a `setEnv` with logging, a sleep, then a `lookupEnv` again results in `Nothing` being printed, followed by another `Nothing` (the second `lookupEnv`), finally followed by the `setEnv` logging.


src/Environment/Portable.hs
```diff
@@ -48,13 +56,14 @@ setEnv key_ value_ overwrite
 
 setEnv_ :: String -> String -> Bool -> IO ()
 setEnv_ key value overwrite =
-  withFilePath key $ \k -> withFilePath value $ \v ->
 #ifdef _WIN32
-    let o = fromBool overwrite
+  withCWString key $ \k -> withCWString value $ \v ->
```
habibalamin: This should be changed back to `withFilePath`.

src/Environment/Portable.hs
```diff
@@ -48,13 +56,14 @@ setEnv key_ value_ overwrite
...
+    let o = fromIntegral $ fromEnum overwrite
     in do
       success <- c_setEnv k v o -- Windows returns non-zero for success.
-      unless (toBool success) $ throwGetLastError "setEnv"
+      unless success $ throwGetLastError "setEnv"
 #else
+  withCString key $ \k -> withCString value $ \v ->
```
habibalamin: This should be changed back to `withFilePath`.

src/Environment/Portable.hs
```diff
@@ -0,0 +1,68 @@
...
+module Environment.Portable (setEnv) where
+
+import Foreign.C
+#ifdef mingw32_HOST_OS
+import GHC.Windows
+import Data.Maybe (isNothing)
+import System.Environment (lookupEnv)
+import Control.Monad (unless)
+#endif
+#ifndef mingw32_HOST_OS
```
habibalamin: `#else`

src/Environment/Portable.hs
```diff
@@ -0,0 +1,68 @@
...
+      withCWString value $ \valueCS ->
+      c_SetEnvironmentVariable keyCS valueCS >>= \success ->
+        unless success $ throwGetLastError "setEnv"
+  else return ()
+#else
+  withFilePath key $ \keyP ->
+    withFilePath value $ \valueP ->
+      throwErrnoIfMinus1_ "setenv" $ c_setenv keyP valueP oCB
+        where
+          oCB = fromIntegral $ fromEnum overwrite
```
habibalamin: This was taken from the official source, but I wonder if `fromBool` will work here. It did on a previous incarnation, and I preferred that form.

habibalamin:
This works on macOS 10.12, but we have 3 failing tests on Windows 7. Two of those failures are the Unicode tests that already fail on previous revisions and weren't affected by this work.

The relevant failure is the new test I added to verify that you can set blank environment variables. It fails on Windows. So, I've managed to actually get `setEnv` on Windows to have an effect, but it removes the environment variable (instead of ignoring it or setting it to the right value) when given an empty string as the value.

It's strange that I was able to set blank environment variables in Ruby, Node.JS, C, and even inherit the new blank environment variables across child processes of different runtimes, but doing it from the FFI doesn't seem to work.

\cc @jsl @CristhianMotoche I'd appreciate more eyes on the code, if you have the time.

src/Environment/Portable.hs
```diff
@@ -0,0 +1,68 @@
...
+    invalidArgIOErr = mkIOError InvalidArgument "setEnv" Nothing Nothing
+
+setEnv_ :: String -> String -> Bool -> IO ()
+setEnv_ key value overwrite = do
+#ifdef mingw32_HOST_OS
+  -- If the user wishes to overwrite or the variable
+  -- doesn't exist.
+  keyDoesntExist <- isNothing <$> lookupEnv key
+
+  if overwrite || keyDoesntExist then
```
habibalamin: `when`

habibalamin:
@mrkkrp @jsl @CristhianMotoche anyone in the team I can ask for help on this particular issue (who maybe works with the FFI a lot)?

habibalamin:
I mentioned in #48 that we're just going to ignore blank variables for Windows, and add a switch to enable it for POSIX.

I'll create a new pull request for that, keep this as is for posterity.

PR #51, opened 2017-06-07
Allow blank variables on UNIX

habibalamin:
Here is the work to allow blank variables on UNIX machines (#48). Let me download and set up a Windows VM to test this on before merging it in.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -18,6 +18,7 @@ import System.Exit (exitWith)
 data Options = Options
   { files    :: [String]
   , overload :: Bool
+  , blanks   :: Bool
   , program  :: String
   , args     :: [String]
   } deriving (Show)
@@ -42,13 +43,17 @@ config = Options
                   <> short 'o'
                   <> help "Specify this flag to override existing variables" )
 
+     <*> switch ( long "allow-empty-values"
+                  <> short 'b'
+                  <> help "Allow blank variables (unix only)" )
+
      <*> argument str (metavar "PROGRAM")
 
      <*> many (argument str (metavar "ARG"))
 
 dotEnv :: MonadIO m => Options -> m ()
 dotEnv opts = liftIO $ do
-  mapM_ (loadFile (overload opts)) (files opts)
+  mapM_ (loadFile (overload opts) (blanks opts)) (files opts)
   code <- system (program opts ++ programArguments)
   exitWith code
   where
--- a/dotenv.cabal
+++ b/dotenv.cabal
@@ -72,6 +72,7 @@ library
                       , Configuration.Dotenv.Parse
                       , Configuration.Dotenv.ParsedVariable
                       , Configuration.Dotenv.Text
+                      , System.Environment.Dotenv
 
   build-depends:         base >=4.6 && <5.0
                        , base-compat >= 0.4
@@ -79,6 +80,9 @@ library
                        , text
                        , transformers >=0.4 && < 0.6
                        , exceptions >= 0.8 && < 0.9
+  if !os(windows)
+    build-depends:
+                       unix >= 2.7 && < 2.8
 
   hs-source-dirs:      src
   ghc-options:         -Wall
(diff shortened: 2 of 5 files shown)


CristhianMotoche:
Thanks a lot @habibalamin for the all the effort that you have put in the project. 
I've review your PR and looks great! I can't wait to have this feature in dotenv. :smile: 
LGTM!

We can wait for @jpvillaisaza to approve your PR.
Thanks one more time. :+1:

habibalamin:
We might need to create a settings type for `load` if we add any more options, but I think that's fine for now.

I've tested this on Windows now, both with `stack test` (making sure that the tests for blank variables aren't run, and that the preprocessing doesn't break the test file for Windows) and running `dotenv` to echo environment variables set in the fixture.

habibalamin:
@jpvillaisaza this work is complete, just so you know.

habibalamin:
@CristhianMotoche, is there anyone else who can approve this PR if @jpvillaisaza is too busy?

jpvillaisaza:
@CristhianMotoche Do the changes here relate to the changes in #43 in any way? (cc @habibalamin)

habibalamin:
It looks like it might conflict, but I don't see how it relates. As far as I can tell, #43 is trying to make it so a `.env.example` file can be used to set required variables, while this makes it so variable values can be blank.

CristhianMotoche:
@habibalamin @jpvillaisaza The changes here are not related to #43 that is a new functionality. That needs some review.

Also, we need a little bit of discussion about this PR. I'll be talking with other team members to see what they think. I'm not sure if this change is approved or not. I promise we would solve this by today. Thanks for your patience.

habibalamin:
@CristhianMotoche bump.

CristhianMotoche:
@habibalamin We're discussing a little bit about this issue. I'll let you know ASAP when we have solved this concern. Once again. Thanks for your patience.

CristhianMotoche:
@habibalamin I'm really sorry for the late response. We've been discussing about this change, because it would bring some changes. We agree that allowing empty env vars wouldn't be necessary and if it is we can do a simple trick to solve that. Like the code that [posted here](https://github.com/stackbuilders/dotenv-hs/pull/7#issuecomment-280757759) or maybe something like: `fromMaybe "" <$> lookupEnv "ENV_VAR"` Also, we've found that the flag (`--blank`) wouldn't have any effect for Windows users and that could be anoying. And adding CPP to restrict libraries, tests and code, it is not the best solution. It could be hard to read and mantein.

The point is that we should be fixing that here. Perhaps, it could be a great idea to start pushing in this [ticket](https://ghc.haskell.org/trac/ghc/ticket/12494) so we can have this functionality in our base level libraries.

For this reasons, I'll close this PR for now. Thanks a lot for your hard work @habibalamin and I hope you still contributing to `dotenv-hs`.

habibalamin:
*Curse you, Microsoft.*

habibalamin:
Sorry, I've been busy again, so haven't had a chance to respond to this properly yet. I took the time to find the commit in the Ruby source that allowed blank environment variables on Windows.

I take your point about this being the wrong place to fix the bug, but I think it might be a good idea to keep this open as long as the original bug is open, just so people know it's a known issue and that its progress is being tracked by us. Just a suggestion.

Anyway, here's where Ruby does it: https://github.com/ruby/ruby/commit/2982c5289210c02120172bf631270858681d031d. It really isn't that difficult. The loop on [line 3147 of the new code](https://github.com/ruby/ruby/commit/2982c5289210c02120172bf631270858681d031d#diff-1f049b4f9a6a39cee9cf18b2dc85f637R3147) is where all the action happens. Trivial, really. I might be able to do this for GHC as well; just gotta get everyone agreed on the course of action, since this is a breaking change.

CristhianMotoche:
Thanks for your suggestion @habibalamin. I'll left one issue open for this change and also I'll update our `README`, so that people know which options they have in order to have empty env vars.

PR #56, opened 2017-07-31
Use megaparsec 6.0

juanpaucar:
Fixes https://github.com/fpco/stackage/issues/2666

### DIFF
--- a/dotenv.cabal
+++ b/dotenv.cabal
@@ -1,5 +1,5 @@
 name:                dotenv
-version:             0.3.4.0
+version:             0.4.0.0
 synopsis:            Loads environment variables from dotenv files
 homepage:            https://github.com/stackbuilders/dotenv-hs
 description:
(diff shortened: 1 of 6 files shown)


spec/Configuration/Dotenv/ParseSpec.hs
```diff
@@ -141,5 +142,5 @@ spec = describe "parse" $ do
   it "doesn't allow more configuration options after a quoted value" $
     parseConfig `shouldFailOn` "foo='bar'baz='buz'"
 
-parseConfig :: String -> Either (ParseError Char Dec) [ParsedVariable]
+parseConfig :: String -> Either (ParseError Char Void) [ParsedVariable]
```
sebashack: @juanpaucar could you just tell me why the type `Dec` changed to `Void`?

juanpaucar: Because as mentioned in he first part of the [documentation](https://hackage.haskell.org/package/megaparsec-6.0.2/docs/Text-Megaparsec.html) I defined a type alias for parser `type Parser = Parsec Void String` and `ParseError` must keep the same type. So, in this case it would be `Void`

dotenv.cabal
```diff
@@ -55,7 +55,7 @@ executable dotenv
                        , base-compat >= 0.4
                        , dotenv >= 0.3.1.0
                        , optparse-applicative >=0.11 && < 0.15
-                       , megaparsec >= 5.0 && < 6.0
+                       , megaparsec >= 5.0 && < 7.0
```
sestrella: @juanpaucar I think you need to increase the lower boundary as well.

dotenv.cabal
```diff
@@ -75,11 +75,14 @@ library
 
   build-depends:         base >=4.6 && <5.0
                        , base-compat >= 0.4
-                       , megaparsec >= 5.0 && < 6.0
+                       , megaparsec >= 5.0 && < 7.0
```
sestrella: @juanpaucar same here

PR #57, opened 2017-08-14
Add dotenv-safe

CristhianMotoche:
This PR will do the same as #43 

### TODO:
- [x] Fix tests suite in Travis CI
- [x] Add tests for dotenv-safe
- [x] Update README
- [x] Update Changelog


CristhianMotoche:

app/Main.hs
```diff
@@ -8,35 +9,50 @@ import Data.Monoid ((<>))
...
 
 config :: Parser Options
 config = Options
      <$> some (strOption (
-                  long "file"
-                  <> short 'f'
-                  <> metavar "FILE"
-                  <> help "File to read for options" ))
+                  long "dotenv"
+                  <> short 'e'
```
juanpaucar: could we keep this flag as `f`? that way if somebody is using the executable only, it would be the same usage unless the user wants the safe feature

app/Main.hs
```diff
@@ -8,35 +9,50 @@ import Data.Monoid ((<>))
...
+                  <> metavar "DOTENV"
+                  <> help "File to read for environmental variables" ))
+
+     <*> many (strOption (
+                  long "dotenv-example"
```
juanpaucar: I think that `dotenv` is not necessary in `dotenv-example`. We could keep simply pass `--example` and would still be clear what is being passed

README.md
```diff
@@ -81,15 +81,50 @@ place to understand the nuances of Dotenv file parsing.
...
+$ dotenv -e mydotenvfile myprogram -- --myflag myargument
+```
+
+or:
+
+```shell
+$ dotenv -e mydotenvfile "myprogram --myflag myargument"
+```
+
+Also, you can use a `-x` to use `dotenv-safe` functionality so that you can have
```
jsl: Can you please add a link to the npm dotenv-safe library so that people have a better idea of what we're talking about?

app/Main.hs
jsl: @juanpaucar I agree. We should avoid changing the interface with this new feature.

app/Main.hs
```diff
@@ -8,35 +9,50 @@ import Data.Monoid ((<>))
...
+                  <> short 'x'
```
jsl: The choice of `-x` for this short flag seems really arbitrary. I understand if we want to avoid overlap, but if `-x` is our only option, I would just not provide a short flag at all. Remember that if users want to make it shorter to execute this program they can always define their own alias in their shell of choice, or create a wrapper script.

spec/Configuration/DotenvSpec.hs
```diff
@@ -52,33 +53,48 @@ spec = do
...
 
+    context "when the .env.example is present" $ do
+      let config = Config ["spec/fixtures/.dotenv"] ["spec/fixtures/.dotenv.example"] False
+
+      context "when the needed env vars are missing" $
+        it "should fail with an error call" $
+          loadFile config `shouldThrow` anyErrorCall
+
+      context "when the needed env vars are not missing" $
+        it "should have succeed loading all the needed env vars" $ do
```
jsl: "should succeed when loading all of the needed env vars"...

juanpaucar:
@CristhianMotoche thanks for following the suggestions :shipit:

Issue #58, opened 2017-08-24
dotenv test failure on stackage


cdornan:
We are seeing this failure and have listed the package under `expected-test-failures`:
```
Configuration.Dotenv.Parse
  parse
    parses unquoted values
    parses values with spaces around equal signs
    parses double-quoted values
    parses single-quoted values
    parses escaped double quotes
    supports CRLF line breaks
    parses empty values
    parses unquoted interpolated values
    parses double-quoted interpolated values
    parses single-quoted interpolated values as literals
    does not parse if line format is incorrect
    expands newlines in quoted strings
    does not parse variables with hyphens in the name
    parses variables with "_" in the name
    parses variables with digits after the first character
    allows vertical spaces after a quoted variable
    does not parse variable names beginning with a digit
    strips unquoted values
    ignores empty lines
    ignores inline comments after unquoted arguments
    ignores inline comments after quoted arguments
    allows "#" in quoted values
    ignores comment lines
    doesn't allow more configuration options after a quoted value
Configuration.Dotenv.Text
  parseFile
    returns variables from a file without changing the environment
    recognizes unicode characters
Configuration.Dotenv
  load
    loads the given list of configuration options to the environment
    preserves existing settings when overload is false
    overrides existing settings when overload is true
  loadFile
    loads the configuration options to the environment from a file
    respects predefined settings when overload is false
    overrides predefined settings when overload is true
    when the .env.example is present
      when the needed env vars are missing
        should fail with an error call FAILED [1]
      when the needed env vars are not missing
        should succeed when loading all of the needed env vars FAILED [2]
  parseFile
    returns variables from a file without changing the environment
    recognizes unicode characters
    recognises environment variables
    recognises previous variables
  onMissingFile
    when target file is present
      loading works as usual
    when target file is missing
      executes supplied handler instead

Failures:

  spec/Configuration/DotenvSpec.hs:82:
  1) Configuration.Dotenv.loadFile, when the .env.example is present, when the needed env vars are missing, should fail with an error call
       uncaught exception: IOException of type NoSuchThing (spec/fixtures/.dotenv.example: openFile: does not exist (No such file or directory))

  spec/Configuration/DotenvSpec.hs:86:
  2) Configuration.Dotenv.loadFile, when the .env.example is present, when the needed env vars are not missing, should succeed when loading all of the needed env vars
[...]

PR #59, opened 2017-08-24
Add spec/fixtures to cabal file

CristhianMotoche:
@juanpaucar Do you think this could solve the issue #58 ?

### DIFF
--- a/dotenv.cabal
+++ b/dotenv.cabal
@@ -1,5 +1,5 @@
 name:                dotenv
-version:             0.5.0.0
+version:             0.5.0.1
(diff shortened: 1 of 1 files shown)


juanpaucar:
@CristhianMotoche I checked the generated distributable and the fixtures are still missing.

CristhianMotoche:
I'll generate a new tar.

juanpaucar:
@CristhianMotoche I checked the generated tar from this branch an it's not including the fixtures. Which seems to be the message on the reported issue

PR #65, opened 2017-09-17
Allow .env file as default, closes #62

### LINKED ISSUE #62: .env as the default file
It would be nice to set `.env` as the default file and make `-f` optional.

Proposed by @sestrella

CristhianMotoche:
@juanpaucar Could you take a look, please? It closes #62

This is a workaround since `value` from `optparser-applicative` doesn't work for `some` or `many`.


CristhianMotoche:
@juanpaucar Ready!

PR #69, opened 2017-10-10
Make a safe wrapper for dotenv

CristhianMotoche:
To solve #41


src/Configuration/Dotenv/Scheme/Parser.hs
```diff
@@ -0,0 +1,37 @@
...
+import Text.Megaparsec
+import Text.Megaparsec.Char
+
+import Configuration.Dotenv.Scheme.Helpers
+import Configuration.Dotenv.Scheme.Types
+
+areParseable
+  :: Config
+  -> [(String, String)]
+  -> Bool
```
sebashack: This would be better if you return a `Either e a` where `Left error` is an error message indicating why you could not parse it.

src/Configuration/Dotenv/Scheme/Parser.hs
```diff
@@ -0,0 +1,37 @@
...
+areParseable config envs =
+  let envsWithConf = mapMatchEnvWithConf config envs
+      canBeParsed (v, t)= v `isParseableAs` t
+   in all canBeParsed envsWithConf
+
+isParseableAs
```
sebashack: Now here you could return a list `[Left error]` so that you can have messages related to the variables that could not be parsed.

src/Configuration/Dotenv/Scheme/Helpers.hs
```diff
@@ -0,0 +1,25 @@
+{-# LANGUAGE RecordWildCards #-}
+
+module Configuration.Dotenv.Scheme.Helpers where
+
+import Data.List
+import Data.Maybe
+
+import Configuration.Dotenv.Scheme.Types
+
+matchEnvWithConf
```
sebashack: Coulsider changing the name of this for something like `matchVarWithType` :)

app/Main.hs
```diff
@@ -25,19 +28,33 @@ data Options = Options
...
   } deriving (Show)
 
+checkEnvTypes
+  :: [(String, String)]
+  -> IO ()
+checkEnvTypes envs = do
+  eitherEnvConf <- decodeFileEither schemeFile
+  case eitherEnvConf of
+    Right envConf ->
+      unless (areParseable envConf envs) (error "The types of you envs don't match")
```
sebashack: Give the user more feedback about the types that did not match.

CristhianMotoche:
@sebashack I think this is ready for another review!

spec/fixtures/.scheme.yml
```diff
@@ -0,0 +1,5 @@
+envs:
```
sebashack: I recommend that instead of duplicating the type name, you declare it once and below it you list the name of those variables with that type.

This is to avoid duplication... :)

CristhianMotoche:
@sebashack It is ready for another review. Could you please take some minutes to take a look? Thanks

README.md
```diff
@@ -83,21 +83,59 @@ sintax $(your_command). For instance:
...
 ```
 $ dotenv "echo $DATABASE"
 postgres://myusername@localhost/database
 ```
 
+### Type checking envs
+Env variables are simple strings. However, they can represent other types like
+integers, booleans, IP addresses, emails, URIs, and so on. We provide an interface
+that performs type checking after loading the envs and before running your application.
+If the type check succeed the application is executed, otherwise you will get an
```
sebashack: "If the type-check succeeded"

README.md
```diff
@@ -83,21 +83,59 @@ sintax $(your_command). For instance:
...
+error with the types that missmatch.
```
sebashack: "the types that mismatch"

README.md
sebashack: You are still missing this typo :)

README.md
sebashack: Still missing typo

README.md
sebashack: Still missing typoe :)

src/Configuration/Dotenv/Scheme/Helpers.hs
```diff
@@ -0,0 +1,26 @@
...
+matchVarWithType
+  :: Config           -- ^ List of EnvConf for variables
+  -> (String, String) -- ^ (Env Name, Env Value)
+  -> (String, EnvType)
+matchVarWithType (Config envConfs) (name, value) =
+  let criteria EnvConf{..} = name `elem` envNames
+      maybeEnvConf = find criteria envConfs
+      pairEnvWithConf EnvConf{..} = (value, envType)
+   in
+   case maybeEnvConf of
```
sebashack: Nice... this way we can catch that the user is declaring all the variables in the dotenv...

CristhianMotoche:
@sebashack Sure, I'll do. Thanks.

CristhianMotoche:
@sebashack Thanks for the review. Well, I'm not sure how we can deal with null envs. I mean, even if we add something to the `scheme.yml` to specify that a variable can be empty we won't be able to set it on the environment. Perhaps, I missunderstand the issue. I think we can discuss about it.

README.md
```diff
@@ -83,21 +83,71 @@ sintax $(your_command). For instance:
...
+
+It is a list of type and envs. So, in this example, `DOTENV` must have a value
+of `true` or `false` otherwise it won't be parsed as a boolean value. And envs
+like `PORT` must be any integer. Currently, we are supporting the following types:
+
+- `bool` - Accepts values `false` or `true`
+- `integer` - Accepts values of possitive integers
+- `text` - Any text
+
+You must set **required** if the env var is going to be type checked or not.
```
sebashack: Perhaps it would be better to say something like this: 

`**require** specifies if the env var is obligatory or not. In case you set it to true but do not provide it, you wil get an exception.`

Perhaps this reflects better the implementation we discussed ... :)

CristhianMotoche:
@sebashack I change the logic here.
Since, we are going to have envs defined in the `dotenvs` and the `scheme.yml`, we need to match those definitions.

In one way, if the envs defined in the `dotenvs` are missing in the `scheme.yml` we should report that. So, that we make sure that all envs defined in the `dotenvs` have their respective configuration in the `scheme.yml`.

In the second way, if the configurations defined in the `scheme.yml` are not related to a required env in the `dotenvs` we should report that, too.

Otherwise, we should start checking the value of the envs with the expected type.

Please, let me know what do you think.

README.md
```diff
@@ -83,21 +83,72 @@ sintax $(your_command). For instance:
...
+like `PORT` must be any integer. Currently, we are supporting the following types:
+
+- `bool` - Accepts values `false` or `true`
+- `integer` - Accepts values of possitive integers
+- `text` - Any text
+
+**require** specifies if the env var is obligatory or not. In case you set it to true
+but do not provide it, you wil get an exception.
+
+**NOTE:** All the envs defined in the .env files **must be defined** in the .scheme.yml file.
```
sebashack: This is only necessary for "required variables" ...

README.md
```diff
@@ -83,21 +83,73 @@ sintax $(your_command). For instance:
...
+that performs type checking after loading the envs and before running your application.
+If the type-check succeeded the application is executed, otherwise you will get an
+error with the types that mismatch.
+
+In order to use this functionality you can use the `loadSafeFile` which takes the same
+configuration value as the `loadFile` function. Also, you need to have a `.scheme.yml`
+in your current directory. This file must have the following structure:
+
+```yaml
+- type: bool
```
sestrella: @CristhianMotoche I'm not sure about this structure. Grouping variables by type look a little bit strange since it doesn't' seem very intuitive. I'm proposing to move the `type` at the `variable_name` level, something like:

```yaml
- name: PORT
  type: integer
  required: true

- name: TOKEN
  type: string
  # required: false (Make it optional, with False as default value perhaps?)
```

Then you should be able to create a data structure like:

```hs
data EnvVariable = EnvVariable
  { envVariableName      :: Text
  , envVariableType        :: EnvVariableType
  , envVariableRequired :: Bool
  }
```

README.md
```diff
@@ -83,21 +83,73 @@ sintax $(your_command). For instance:
...
+like `PORT` must be any integer. Currently, we are supporting the following types:
+
+- `bool` - Accepts values `false` or `true`
+- `integer` - Accepts values of possitive integers
+- `text` - Any text
+
+**require** specifies if the env var is obligatory or not. In case you set it to true
+but do not provide it, you wil get an exception.
+
+**NOTE:** All the variables which are **required** in the `scheme.yml` must be defined
```
sestrella: @CristhianMotoche `schema.yml` perhaps? instead of `scheme.yml`

app/Main.hs
```diff
@@ -70,3 +73,6 @@ config = Options
 
      <*> many (argument str (metavar "ARG"))
 
+     <*> switch ( long "safe"
```
sestrella: @CristhianMotoche I would like to suggest a few changes here:

- Turn safe mode on automatically when the schema file is present.
- Add an option `--schema` to override the default path of the schema file.
- Add a flag `--no-type-check` to turn safe mode off explicitly.

What do you think?

README.md
CristhianMotoche: @sestrella the current structure will avoid duplication, since we will only declare the type once.
WDYT?

README.md
CristhianMotoche: @sestrella Sure, no problem.

app/Main.hs
CristhianMotoche: @sestrella I agree. Perhaps, we can change `--no-type-check` with `--no-schema`.

README.md
sestrella: @CristhianMotoche you could achieve the same using plain YAML:

```yaml
- &numeric { type: integer, required: true }

- name: HOST
  <<: *numeric

- name: DB
  <<: *numeric
```

http://yaml.org/type/merge.html

app/Main.hs
sestrella: @CristhianMotoche sounds good :+1:

README.md
CristhianMotoche: 😲 🆒 @sestrella I'll change it then.

PR #73, opened 2017-12-20
Add .scheme.yml to .cabal file

CristhianMotoche:
This file was missing in the construction of the sdist file.


CristhianMotoche:
@juanpaucar I haven't uploaded the package, so I think it won't be necessary. WDYT?

juanpaucar:
`s/0.5.2.1/0.5.2.0/g` could be included here and then 🚢 -it

PR #74, opened 2017-12-20
Changes in the safe wrapper

CristhianMotoche:
# TODO:
- [x] Change structure of the expected file
- [x] Run safe wrapper when the `schema.yml` is present or the `--schema` is set.
- [x] Add flag `--schema`
- [x] Add flag `--no-schema`
- [x] Update `README`

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,22 +99,21 @@ If the type-check succeeded the application is executed, otherwise you will get 
 error with the types that mismatch.
 
 In order to use this functionality you can use the `loadSafeFile` which takes the same
-configuration value as the `loadFile` function. Also, you need to have a `.scheme.yml`
+configuration value as the `loadFile` function. Also, you need to have a `.schema.yml`
 in your current directory. This file must have the following structure:
 
 ```yaml
-- type: bool
-  envs:
-    - name: DOTENV
-      required: true
-    - name: OTHERENV
-      required: false
-- type: integer
-  envs:
-    - name: PORT
-      required: true
-    - name: TOKEN
-      required: true
+- name: DOTENV
+  type: bool
+  required: true
+- name: OTHERENV
+  type: bool
+- name: PORT
+  type: integer
+  required: true
+- name: TOKEN
+  type: text
+  required: false
 ```
 
 It is a list of type and envs. So, in this example, `DOTENV` must have a value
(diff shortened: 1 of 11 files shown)


CristhianMotoche:
@sestrella Could you take a look, please?

src/Configuration/Dotenv/Scheme/Types.hs
```diff
@@ -21,33 +23,17 @@ instance FromJSON EnvType where
...
     , required :: Bool
     } deriving (Show, Eq)
 
 instance FromJSON Env where
   parseJSON (Object m) =
     Env
       <$> m .: "name"
-      <*> m .: "required"
+      <*> m .: "type"
+      <*> (fromMaybe False <$> m .:? "required")
```
sestrella: @CristhianMotoche what about:

```hs
m .:? "required" .!= False
```

There is already a function provided by [aeson](http://hackage.haskell.org/package/aeson-1.2.3.0/docs/Data-Aeson.html#v:.-33--61-)

README.md
```diff
@@ -99,22 +99,21 @@ If the type-check succeeded the application is executed, otherwise you will get
...
-      required: true
-    - name: OTHERENV
-      required: false
-- type: integer
-  envs:
-    - name: PORT
-      required: true
-    - name: TOKEN
-      required: true
+- name: DOTENV
```
sestrella: @CristhianMotoche is there any check to prevent `EnvVaribles` duplication?

```yaml
- name: PORT
  type: bool
- name: PORT
  type: numeric
```

src/Configuration/Dotenv/Scheme/Types.hs
```diff
@@ -21,33 +23,17 @@ instance FromJSON EnvType where
...
+    , envType  :: EnvType
     , required :: Bool
     } deriving (Show, Eq)
 
 instance FromJSON Env where
   parseJSON (Object m) =
     Env
       <$> m .: "name"
-      <*> m .: "required"
+      <*> m .: "type"
```
sestrella: @CristhianMotoche since most likely `text` would be the most common type. What do you think about setting a default type?

```hs
m .:? "type" .!= EnvText
```

CristhianMotoche: @sestrella I think even then, the user should _explicitly_ specify the type.

README.md
CristhianMotoche: @sestrella Currently, we don't have. However, I was checking that the system has the current behaviour: it matches the types with the value and then type checks for all the types specified.

For instance, in your example the value of `PORT` will be evaluated as a `bool` and `numeric`. The result from that will always fail. I think, this could be enough. WDYT?

sestrella: @CristhianMotoche do you mind adding an issue, so we could tackle that later, please?

src/Configuration/Dotenv/Scheme/Types.hs
sestrella: @CristhianMotoche fair enough

PR #76, opened 2017-12-27
Fix module description markup (Configuration.Dotenv.Types)

jpvillaisaza:
The current module description markup for `Configuration.Dotenv.Types` uses both `{- |` and `--`, which yields:

screen shot 2017-12-27 at 4 32 24 pm

Instead of (using this change):

screen shot 2017-12-27 at 4 33 05 pm


CristhianMotoche:
@jpvillaisaza LGTM! Thanks

PR #78, opened 2018-01-05
Fix mismatch of env variable names in test

laMudri:
This test was failing, and now succeeds. I'm guessing this was the intention, because I don't see `USER` mentioned anywhere else.

### DIFF
--- a/spec/Configuration/DotenvSpec.hs
+++ b/spec/Configuration/DotenvSpec.hs
@@ -86,7 +86,7 @@ spec = do
       context "when the needed env vars are not missing" $
         it "should succeed when loading all of the needed env vars" $ do
           setEnv "ANOTHER_ENV" "hello"
-          me <- getEnv "USER"
+          me <- getEnv "ME"
           home <- getEnv "HOME"
           loadFile config `shouldReturn` [("DOTENV","true"),("UNICODE_TEST","Manab\237"),("ENVIRONMENT", home),("PREVIOUS","true"),("ME", me),("ANOTHER_ENV","")]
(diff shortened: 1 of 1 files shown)


CristhianMotoche:
Hello @laMudri 
You made me realize and issue with our tests. Thanks!

Let me explain to you:

The configuration for `loadFile` is defined [here](https://github.com/laMudri/dotenv-hs/blob/e50cfae90d8b890ba0e6889b5af2b1c0eb58cc96/spec/Configuration/DotenvSpec.hs#L80) which is  the same file for other tests. As you can see, we load the configuration from `spec/fixtures/.dotenv` for some tests. Thus, some tests before the one in which you did the change set the env var `ME`, and that is why the test still working.

If you remove the other tests and only run that one (clearing the env var `ME` you will notice that the test fails). That won't happen if we get the env `USER` which is an env var that is [set when you log in](http://man7.org/linux/man-pages/man1/login.1.html) in most unix like.

So, what is the issue with our tests, we are not clearing all the envs defined in the `spec/fixtures/.dotenv`  and that could cause other issues later.

So, I'll close this PR for now and I'll add an issue to the bug tracker.

Thanks @laMudri

laMudri:
I'm not sure, but I think relying on environment variables like this will break the tests when they're run in a pure environment. I was trying to build this in Nix, and the error was that `USER` was not found in the environment (which makes sense, because a build shouldn't be able to depend on the user).

CristhianMotoche:
@laMudri I see your point and I think we can fix this by setting the necessary envs `before` running the tests.

Issue #79, opened 2018-01-05
Clear all the envs defined in `.spec/fixtures/.dotenv` after the tests execution


CristhianMotoche:
We should clear the envs set in environment after every test execution, so every single test execution is isolated. Otherwise, we could have issues like the one described in #78

PR #80, opened 2018-01-05
Clear dotenvs after every test execution, solve #79

CristhianMotoche:
@juanpaucar Could you take a look, please? To solve #79

### DIFF
--- a/spec/Configuration/DotenvSpec.hs
+++ b/spec/Configuration/DotenvSpec.hs
@@ -54,7 +54,7 @@ spec = do
 
       lookupEnv "foo" `shouldReturn` Just "new setting"
 
-  describe "loadFile" $ after_ (mapM_ unsetEnv ["DOTENV" , "ANOTHER_ENV"]) $ do
+  describe "loadFile" $ after_ (mapM_ unsetEnv dotenvFixtures) $ do
     it "loads the configuration options to the environment from a file" $ do
       lookupEnv "DOTENV" `shouldReturn` Nothing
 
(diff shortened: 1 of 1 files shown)


spec/Configuration/DotenvSpec.hs
```diff
@@ -120,13 +120,22 @@ spec = do
       liftM (!! 4) (parseFile "spec/fixtures/.dotenv") `shouldReturn`
         ("ME", me)
 
-  describe "onMissingFile" $ after_ (unsetEnv "DOTENV") $ do
+  describe "onMissingFile" $ after_ (mapM_ unsetEnv dotenvFixtures) $ do
```
juanpaucar: why not extract this to a function `mapM_ unsetEnv dotenvFixtures` too?

CristhianMotoche:
@juanpaucar I change the perspective of the PR. Now, I clean the entire environment until every test execution. For the `loadFile` tests, I clean the environment and then set up the necessary envs. To get the username, I run the command `whoami` and extract the text result. Then, I compare the result text with the expected result, which should be the same. WDYT?

juanpaucar (approved):
LGTM thanks

CristhianMotoche:
Thanks to you @juanpaucar

sternenseemann:
This can break the test suite completely since it unsets `PATH` which may lead to `shell "whoami"` which is used in a couple of places may fail. Probably the simplest alternative is to use [getLoginName](https://hackage-content.haskell.org/package/unix-2.8.7.0/docs/System-Posix-User.html#v:getLoginName) since it seems to assume Unix anyways.

PR #86, opened 2018-02-28
Don't allow duplicated values in the schema configuration

### LINKED ISSUE #75: Don't allow more than one definition in the Scheme for the same env
Currently, the user can define multiple configurations for the same env. However, the system has the following behaviour: it matches the types with the value and then type checks for all the types specified.

For instance, 
```yml
- name: PORT
  type: integer
  required: true
- name: PORT
  type: bool
  required: true
```

`PORT` will be evaluated as a bool and integer, then dotenv will always fail.
Although, it would be better to display an error message.

dbalseiro:
Fixes #75

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ install:
  - cabal --version
  - echo "$(ghc --version) [$(ghc --print-project-git-commit-id 2> /dev/null || echo '?')]"
  - travis_retry cabal update
- - cabal install --only-dependencies --enable-tests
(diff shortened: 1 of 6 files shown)


CristhianMotoche:
@dbalseiro It looks good so far. However, I think we are using `error` many times in the application, and after reading a little bit about [error and exception handling](http://book.realworldhaskell.org/read/error-handling.html) it would be better to start using an [specific exception](http://hackage.haskell.org/package/base-4.10.1.0/docs/Control-Exception-Base.html#t:Exception) for dotenv. Please, let me know what you think.

PR #96, opened 2018-05-23
Improve dotenv safe

CristhianMotoche:
Now, we are passing a `ValidatorMap` which is just a type alias for a `Map Text (Text -> Bool)` to `loadSafeDotenv`


CristhianMotoche:

src/Configuration/Dotenv/Scheme/Types.hs
```diff
@@ -8,32 +8,43 @@ import Control.Applicative ((<*>), pure)
...
 
+import Data.Map.Lazy (Map)
+
 import Data.Yaml
 
-data EnvType =
-  EnvInteger
-    | EnvBool
-    | EnvText
+import Data.Text
```
juanpaucar: Could import only `Text`?

CristhianMotoche:
@juanpaucar I'm changing the export module of `loadSafeFile`. Additionally, I'm exporting the `Configuration.Dotenv.Types` module from `Configuration.Dotenv`. Therefore, I consider this a major change. Please, let me know if you agree with it.

Issue #88, opened 2018-03-04
Handle exceptions


CristhianMotoche:
I think we are using many time `error` in the application. However, I don't think that is the correct way to handle some errors. Perhaps, we should start using an [specific exception](http://hackage.haskell.org/package/base-4.10.1.0/docs/Control-Exception-Base.html#t:Exception) type.

po5i:
![](https://github.trello.services/images/mini-trello-icon.png) [Handle exceptions](https://trello.com/c/gq1OnjDo/37-handle-exceptions)

CristhianMotoche:
I think it make sense to use `error` to stop a program's execution for now. Exceptions shouldn't be necessary in this case. So, I'll close this issue for now.

PR #97, opened 2018-08-23
Rename Scheme with Schema

### LINKED ISSUE #87: Rename scheme with schema
@dbalseiro Suggest that we should rename the word `scheme` with `schema` which would be a better name.

CristhianMotoche:
Fixes #87


CristhianMotoche:
I think renaming make sense since `scheme` can be a [verb or a noun](https://www.merriam-webster.com/dictionary/scheme). However, I don't think it's worth to increase the version for a renaming right now. I'll close this for now.

CristhianMotoche:
Closed due to: https://github.com/stackbuilders/dotenv-hs/issues/88#issuecomment-580247315

PR #103, opened 2018-12-04
Allow blank variables

habibalamin:
#48.

I have a bad feeling this will break for GHC 7.10, since I removed compat CPP macros. I had to update the lower bound on base — not sure if this is acceptable — to get consistent behaviour here.

Just gonna wait for CI.


habibalamin:
Huh… looks like only GHC 8.4 support `base-4.11.0.0`.

CristhianMotoche:
@habibalamin Yeah. I see Stack LTS 12 works with `GHC-8.4.3` so it should work with `base-4.11.0.0`. I suppose we'll still have to depend on CPP for backward compatibility. I suppose we can clarify in our `README` that `System.Environment.Blank` is used to set environmental variables **only** for `base >= 4.11`.
What do you think?


I'll ask internally what they think about updating the bounds of `base`.

juanpaucar:
Here are my two cents: According to Haskell Survey, this is the distribution for GHC versions usage. Dropping the support would be dropping 46% of possible users. I would commend to use pragmas edpending on the version. But we can a further discussion here about what could be more "pragmatic" 

screenshot from 2018-12-05 09-54-08

habibalamin:
This is not ready for merge yet.

I need to write some extra tests for blank variables on 8.4.4 and also factor out a `Dotenv.Environment` module to avoid repeating certain CPP macros.

I'm just taking advantage of the CI infrastructure.

juanpaucar:
Let us know when it's ready to remove the label

habibalamin:
@juanpaucar done. If CI is green for all versions, I'm happy with that.

dotenv.cabal
```diff
@@ -78,6 +78,7 @@ executable dotenv
 
 library
   exposed-modules:      Configuration.Dotenv
+                      , Configuration.Dotenv.Environment
```
CristhianMotoche: @habibalamin Is it necessary to expose this module?

CristhianMotoche:
@habibalamin Looks great! Thanks a lot for your contributions. I just left one comment. Let me know what you think.

CristhianMotoche:
@habibalamin Could you please update dotenv's version so that we can upload the new version to hackage?

dotenv.cabal
habibalamin: It’s not necessary, but I thought a module which wraps the CPP stuff we have to do would be generally useful for many other people.

For example, at certain points, they may themselves want to write their own functions which wrap `parseFile`, so they can write their own `loadFile`, just like I do [here](https://github.com/stackbuilders/dotenv-hs/pull/7#issuecomment-280757759) before the blank environment variable issue was fixed.

Of course, if they require blank environment variables, like I do, then they’ll just have to use the `System.Environment.Blank`, but if they don’t care about that, but do want to wrap `parseFile` for other reasons, they’d probably want to use the same environment functions we use (for consistency with unwrapped behaviour by default), which are automatically selected based on version of `base`, instead of writing their own CPP stuff like we did.

CristhianMotoche: Ok, I think you have a point @habibalamin .  Let's leave that module exposed.

habibalamin:
@CristhianMotoche done.

habibalamin:
@CristhianMotoche just a reminder that this is ready to merge.

Issue #105, opened 2019-07-05
Support for GHC 8.6


CristhianMotoche:
# TODO:
- [ ] Make sure dotenv works with GHC 8.6
- [ ] Add the version 8.6 to the `.travis.yml` file

ibarrae:
Mind if I take this one @CristhianMotoche?

CristhianMotoche:
Go ahead @ibarrae Thanks.

CristhianMotoche:
In favor of #106

PR #106, opened 2019-07-18
Add support for GHC 8.6

ibarrae:
Solves #105.

LTS 13.28 was released a few days ago and it also contains all packages marked before as `extra-deps` on `stack.yaml`.


CristhianMotoche:
Thanks @ibarrae  It's good to know that it didn't broke.

Issue #112, opened 2019-11-13
Create releases with pre-compiled binaries


sestrella:
`dotenv` is meant to be used as an executable, therefore, in order to simplify the installation process, we could provide pre-compiled binaries so end-users could just download and copy the binary instead of installing the whole package via `stack` or `cabal`.

**Documentation**
- https://help.github.com/en/github/administering-a-repository/creating-releases
- https://docs.travis-ci.com/user/deployment/releases/

**References**
- https://github.com/commercialhaskell/stack/releases

CristhianMotoche:
I've never done this before. I'd love to take this.

CristhianMotoche:
Hey @sestrella As we discussed, let's do this step incrementally. I think our final goal is to make `dotenv` binary work for the following operating systems:
- [ ] Linux x86_64
- [ ] Mac osX
- [ ] Windows

I'll keep you updated with any progress.

sestrella:
@CristhianMotoche sounds good to me, thank you for carrying this on.

CristhianMotoche:
@sestrella I'm trying to set the release notes. Here is my first attempt: https://github.com/stackbuilders/dotenv-hs/releases/tag/untagged-8972a8d79f11a59c973f

Two things here:

1) The advanced options of [travis deploy didn't work](https://docs.travis-ci.com/user/deployment/releases/#advanced-options), because [formatting the `body` isn't supported](https://github.com/travis-ci/dpl/issues/155).

2) So, based on this [issue](https://github.com/travis-ci/dpl/issues/155), I followed the suggestion described [here](https://github.com/travis-ci/dpl/issues/155#issuecomment-500130415).

Let me know what you think.

CristhianMotoche:
Travis has the option to build on Mac OS and even Windows, I think. However, it cannot collect artifacts after builds. Info [here](https://docs.travis-ci.com/user/build-stages/#data-persistence-between-stages-and-jobs). An option that Travis suggests is to set up a S3 bucket and push the artifacts there. And later, move those binaries from S3 to a GitHub release.

What do you think about that approach? /cc @sestrella @juanpaucar

sestrella:
@CristhianMotoche nice suggestion, in fact, that seems to be the way to go in some well-know packages such as [stack](https://github.com/commercialhaskell/stack/tree/master/etc/scripts). I think adding some deployment scripts - to build the binaries locally - would be a good idea.

PR #114, opened 2019-11-18
Create static binary linux 64

CristhianMotoche:
Solves #112 partially.

## Description
I configured travis to create a draft release on each merge to a branch (I suspect that will be `master`). 

Resulting draft release: https://github.com/stackbuilders/dotenv-hs/releases/tag/untagged-102599435dcc9f6e17c9

**Note:** I just need to change the branch name, currently it's deploying when something is merged in this branch for testing).

### TODO:
- [ ] Change the Github API key (it's using mine now)
- [x] Change `create-static-binary-linux-64` to `master`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,17 +10,16 @@ ghc:
 
 before_deploy: |
   if [[ $TRAVIS_HASKELL_VERSION == "8.6.5" ]]; then
+    tar -C dist/build/dotenv -cvf dotenv-linux-x86_64-static.tar.gz dotenv
     cabal install -f 'static'
-    tar -C dist/build/dotenv -cvf dotenv.static.linux.x86_64.tar.gz dotenv
   fi
 
 deploy:
   provider: releases
   api_key: "$GITHUB_TRAVIS_DEPLOY"
-  file: "dotenv.static.linux.x86_64.tar.gz"
+  file: "dotenv-linux-x86_64-static.tar.gz"
   skip_cleanup: true
   draft: true
   on:
-    branch: create-static-binary-linux-64
     condition: $TRAVIS_HASKELL_VERSION == "8.6.5"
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -7,3 +7,19 @@ ghc:
   - "8.2"
   - "8.0"
   - "7.10"
+
+before_deploy: |
+  if [[ $TRAVIS_HASKELL_VERSION == "8.6.5" ]]; then
+    tar -C dist/build/dotenv -cvf dotenv-linux-x86_64-static.tar.gz dotenv
+    cabal install -f 'static'
```
sestrella: @CristhianMotoche perhaps I'm missing something here but, I think you need to compile the package again with` -f static` on in order to produce a static binary.

.travis.yml
```diff
@@ -7,3 +7,19 @@ ghc:
...
+  fi
+
+deploy:
+  provider: releases
+  api_key: "$GITHUB_TRAVIS_DEPLOY"
+  file: "dotenv-linux-x86_64-static.tar.gz"
+  skip_cleanup: true
+  draft: true
```
sestrella: @CristhianMotoche draft releases need to be manually approved, any particular reason why we want to add an extra step here?

.travis.yml
CristhianMotoche: Hi @sestrella I think i accidentally switched lines 13 and 14. BTW, I'm not sure if we need to build the whole package again with `-f static`. I was testing locally and doing a `cabal install` produces the static binary.

.travis.yml
CristhianMotoche: Hey @sestrella I think with the draft release, we're able to manually add release notes (e.g. the latest changes), test that the binary works correctly before doing a real release.

PR #115, opened 2019-11-20
Format release notes with ChangeLog content

CristhianMotoche:
Solves part of #112 
## Description

I'm using `github-release-cli` to deploy the releases.

Example of draft release generated with this feature: https://github.com/stackbuilders/dotenv-hs/releases/tag/untagged-e5b89fbe11e4932ef8a5

## NOTES
We still need to remove the extra notes from previous releases form the ChangeLog.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ deploy:
   file: "dotenv-linux-x86_64-static.tar.gz"
   skip_cleanup: true
   release_notes_file: "CHANGELOG.md"
+  edge: true
   draft: true
   on:
     branch: release-with-notes
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,5 +24,6 @@ deploy:
   script: ./github-releaser -draft -verbose
   skip_cleanup: true
   on:
+    branch: release-with-notes
     condition: $TRAVIS_HASKELL_VERSION == "8.6.5"
-    tags: true
+    tags: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ before_deploy: |
     tar -C dist/build/dotenv -cvf dotenv-linux-x86_64-static.tar.gz dotenv
     curl https://github.com/lindell/github-release-cli/releases/download/1.3.0/github-releaser-travis -L --output github-releaser && chmod +x github-releaser
     export GITHUB_OAUTH_TOKEN="$GITHUB_TRAVIS_DEPLOY"
-    export RELEASE_NAME=Dotenv-Release-v$(cabal exec dotenv -- --version)
+    export RELEASE_NAME=v$(cabal exec dotenv -- --version)
     export BODY=$(envsubst < ./CHANGELOG.md)
     export FILES=dotenv-linux-x86_64-static.tar.gz
   fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,14 +12,17 @@ before_deploy: |
   if [[ $TRAVIS_HASKELL_VERSION == "8.6.5" ]]; then
     cabal install -f 'static'
     tar -C dist/build/dotenv -cvf dotenv-linux-x86_64-static.tar.gz dotenv
+    curl https://github.com/lindell/github-release-cli/releases/download/1.3.0/github-releaser-travis -L --output github-releaser && chmod +x github-releaser
+    export GITHUB_OAUTH_TOKEN="$GITHUB_TRAVIS_DEPLOY"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -12,14 +12,18 @@ before_deploy: |
   if [[ $TRAVIS_HASKELL_VERSION == "8.6.5" ]]; then
     cabal install -f 'static'
     tar -C dist/build/dotenv -cvf dotenv-linux-x86_64-static.tar.gz dotenv
+    curl https://github.com/lindell/github-release-cli/releases/download/1.3.0/github-releaser-travis -L --output github-releaser && chmod +x github-releaser
+    export GITHUB_OAUTH_TOKEN="$GITHUB_TRAVIS_DEPLOY"
+    export RELEASE_NAME=Dotenv-Release-v$(cabal exec dotenv -- --version)
+    export BODY=$(envsubst < ./CHANGELOG.md)
```
sestrella: @CristhianMotoche I'mnot sure about adding the whole `CHANGELOG` to each release, it might too verbose, perhaps we only need to include the latest changes.

.travis.yml
```diff
@@ -12,14 +12,18 @@ before_deploy: |
   if [[ $TRAVIS_HASKELL_VERSION == "8.6.5" ]]; then
     cabal install -f 'static'
     tar -C dist/build/dotenv -cvf dotenv-linux-x86_64-static.tar.gz dotenv
+    curl https://github.com/lindell/github-release-cli/releases/download/1.3.0/github-releaser-travis -L --output github-releaser && chmod +x github-releaser
+    export GITHUB_OAUTH_TOKEN="$GITHUB_TRAVIS_DEPLOY"
+    export RELEASE_NAME=Dotenv-Release-v$(cabal exec dotenv -- --version)
```
sestrella: @CristhianMotoche what do you think about dropping the prefix here `Dotenv-Release-`, taking into account all releases are uploaded to the same repository, I would say there is no need to add the project name to the release, perhaps just the version?

.travis.yml
```diff
@@ -12,14 +12,18 @@ before_deploy: |
   if [[ $TRAVIS_HASKELL_VERSION == "8.6.5" ]]; then
     cabal install -f 'static'
     tar -C dist/build/dotenv -cvf dotenv-linux-x86_64-static.tar.gz dotenv
+    curl https://github.com/lindell/github-release-cli/releases/download/1.3.0/github-releaser-travis -L --output github-releaser && chmod +x github-releaser
```
sestrella: @CristhianMotoche I'm not against using a third-party tool here, but I think `curl` could get the job done. My concern about bringing up another tool is that we have to take care of keeping it up to date and install all required dependencies, compared to `curl` which is a system dependency that could be easily updated.

.travis.yml
```diff
@@ -12,14 +12,18 @@ before_deploy: |
...
+    export BODY=$(envsubst < ./CHANGELOG.md)
+    export FILES=dotenv-linux-x86_64-static.tar.gz
   fi
 
 deploy:
-  provider: releases
-  api_key: "$GITHUB_TRAVIS_DEPLOY"
-  file: "dotenv-linux-x86_64-static.tar.gz"
+  provider: script
+  script: ./github-releaser -draft -verbose
```
sestrella: @CristhianMotoche taking into account some this script depends on all the operations done at `before_deploy`, maybe we should put everything in one place? That way we also avoid polluting code with conditionals:

```
# file deploy.sh
curl -H "Authorization: token $GITHUB_TOKEN" https://api.github.com/repos/$TRAVIS_REPO_SLUG/releases
```

References: https://docs.travis-ci.com/user/environment-variables/#default-environment-variables

.travis.yml
CristhianMotoche: You're right @sestrella Let me add some script to filter the change log.

.travis.yml
CristhianMotoche: Oh, make total sense @sestrella :+1:

.travis.yml
CristhianMotoche: Make sense @sestrella Let me take a look at the GitHub API for that.

CristhianMotoche:
Hi @sestrella @juanpaucar  I'm retaking this PR. Sorry for taking so long. :cry:
Well, I found that there is an easy way to deploy release notes to the draft releases after taking a look at this long [thread](https://github.com/travis-ci/dpl/issues/155).

You can take a look at the draft release [here](https://github.com/stackbuilders/dotenv-hs/releases/tag/untagged-12d70da30a94b55fd989). Please, let me know if you have any comment.

PR #116, opened 2020-01-06
Update license date range

CristhianMotoche:
I set the license as a date range as we have in other [projects](https://github.com/stackbuilders/hapistrano/blob/master/LICENSE#L1). Also, the start date was in `2015` as this [commit](https://github.com/stackbuilders/dotenv-hs/commit/0f49e3e6c711c3166ef9215f224bde1abaae7524#diff-9879d6db96fd29134fc802214163b95aR1) shows.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2016 Justin Leitgeb
+Copyright (c) 2015-2020 Justin Leitgeb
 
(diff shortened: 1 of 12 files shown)


juanpaucar:
@CristhianMotoche Could you also update this one, please?
https://github.com/stackbuilders/dotenv-hs/blob/master/dotenv.cabal#L39

Issue #121, opened 2020-04-29
dotenv picks the first instead of the last value


rudymatela:
See:

```
$ cat > my_script.hs
#!/bin/bash
echo $FOO
echo $GOO
^D
```

```
$ cat > .env
#!/bin/sh
FOO=first
GOO=second
FOO=third
^D
```

```
$ dotenv -f.env ./my_script.hs 
first
second
```

I would expect for the result to be actually:
```
third
second
```

Now, that may be just a personal preference...  However I justify it as it is less surprising for `dotenv` files to behave like shell scripts and have a preference for the last set value:

```
$ source .env
$ echo $FOO
third
$ echo $GOO
second
```

rudymatela:
I was wondering if Python's `dotenv` behaved in the same way.  Looks like Python's `dotenv` is different from Haskell's but equal to bashes `source`:

CristhianMotoche:
Hello @rudymatela This issue and your PR (#122) have been open for a very long time. Let's try to wrap it up.

I like the idea to have tools working in a similar way and I would like to see how other dotenv tools behave. I'll ask a member in my team to take a look at this and report their findings.

CristhianMotoche:
Hello! 
@anddriex is going to give a look at this issue to replicate something similar to what other dotenv tools are doing.

CristhianMotoche:
Solved by #161  thanks for the contribution @anddriex and thankd @rudymatela for reporting it!

PR #122, opened 2020-04-29
pick last instead of first value

rudymatela:
This makes it so that we prefer latest than earlier occurrences of variable
settings.

Before the change
-----------------

    $ cat > my_script.hs
    #!/bin/bash
    echo $FOO
    echo $GOO
    ^D

    $ cat > .env
    #!/bin/sh
    FOO=first
    GOO=second
    FOO=third
    ^D

    $ dotenv -f.env ./my_script.hs
    first
    second

After the change
----------------

    $ cat > my_script.hs
    #!/bin/bash
    echo $FOO
    echo $GOO
    ^D

    $ cat > .env
    #!/bin/sh
    FOO=first
    GOO=second
    FOO=third
    ^D

    $ dotenv -f.env ./my_script.hs
    third
    second

This makes dotenv-hs consistent with both Python's dotenv and Bash/Sh's source.

The implementation may look innefficient with

    reverse . nubBy (...) . reverse

however it is as efficient as the surrounding code which is _O(n^2)_ overall as
Data.List.union is _O(n^2)_ already.

The ultimate reverse may not be needed, but I have kept it so the order in
which variables are set is unchanged.

This fixes issue #121.

### DIFF
--- a/src/Configuration/Dotenv.hs
+++ b/src/Configuration/Dotenv.hs
@@ -38,8 +38,9 @@ import           Configuration.Dotenv.Types          (Config (..),
 import           Control.Monad                       (liftM, when)
 import           Control.Monad.Catch
 import           Control.Monad.IO.Class              (MonadIO (..))
+import           Data.Function                       (on)
 import           Data.List                           (intersectBy, union,
-                                                      unionBy)
+                                                      unionBy, nubBy)
 import           System.Directory                    (doesFileExist)
 import           System.IO.Error                     (isDoesNotExistError)
 import           Text.Megaparsec                     (errorBundlePretty, parse)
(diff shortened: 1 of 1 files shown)


rudymatela:
Needless to say: this may or may not be desirable.  One can argue that preferring the _last_ occurence instead of the _first_ is a personal preference.  I do prefer pciking the last one for consistency with other tools.

So feel free to reject this PR if you disagree.  :-)

CristhianMotoche:
Thanks for the contribution @rudymatela :smile: 

If other dotenv tools work this way, I suppose that's the _de facto standard_, we should follow it as well. Although, I think this could be a breaking change in case someone is already expecting dotenv to set the _first_ ocurrence rather than the _last_ one.

I'd like to hear your thoughts @jsl @sestrella @juanpaucar

juanpaucar:
In personal preference with `.bashrc` or `.zshrc` configurations I have always come to expect that the latest value overrides previous values. If the python equivalent is using that approach, then, I find it even more desirable to go on that way

sestrella:
@rudymatela Thank you for submitting this patch, I wondering if duplicate keys in a `.env` value is a use case or a user input error? If it is an error, maybe we should report this back to the user?

CristhianMotoche:
Hello @rudymatela I really appreciate your contribution to the dotenv-hs library. Your solution is simple and straightforward. However, I prefer to go with the change done in #161 since it follows an approach similar to the one from other libraries from other languages.

I'm going to prepare a release with this change so that I can close #121 and this PR, which have been open for a very long time now.

Once again, thanks a lot for using and contributing to the library. I really appreciate your contributions and any future PR is very welcome.

CristhianMotoche:
One more time, thanks for the contribution @rudymatela The last dotenv release (0.11.0.0) contains the fix for this. So, I'll close this PR.

PR #125, opened 2020-09-10
Allow megaparsec 9.0

CristhianMotoche:


dotenv.cabal
```diff
@@ -100,7 +100,7 @@ library
   build-depends:         base >=4.7 && <5.0
                        , base-compat >= 0.4
                        , directory
-                       , megaparsec >= 7.0.1 && < 9.0
+                       , megaparsec >= 7.0.1 && < 9.1
```
juanpaucar, line 103: Would it be better to enable it for `< 10`? So that you don't have to pump version dependencies for nonbreaking changes

sestrella:
@CristhianMotoche I wonder if you could force `megaparsec-9.0.0` temporarily just to make sure it works properly:

```yaml
# stack.yaml
...
extra-deps:
  - megaparsec-9.0.0
```

CristhianMotoche:
Done @sestrella It went well

Issue #128, opened 2021-06-02
Move from Travis CI to GH actions


fefi95:
We want to move from Travis CI to GH actions due to a change in its policies. The configuration for GH Actions should cover the same that Travis does, that is:

- [x] Common check on Linux. Build and tests for versions: `8.10,  8.8,  8.6,  8.4,  8.2, 8.0, 7.10`
- [x] Common check on OSX. Build and tests for versions: `8.10`
- [x] Bundle the project and create a release in GH based on a given tag.

CristhianMotoche:
Thanks a lot @fefi95 for the contributions. I'm closing this issue now.

PR #129, opened 2021-06-02
Setup GH actions to build and test the project on Linux

fefi95:
# Issue: #128 

##  Description
This PR sets up GH actions to build and test the project on Linux and MacOS and removes Travis CI. The release workflow is not set up yet and will be addressed in another PR

> Note: Travis is using the cabal version 2.4, I tried upgrading to 3.4 but had issues building the project. I'll open another issue to update the version


CristhianMotoche:
Hey @fefi95 Looks good! Thanks for the contribution. Could you please run the GH actions again to make sure the cache is saved and restored?

fefi95:
Hi @CristhianMotoche, Thanks for taking a look! Here are 2 runs of the CI, one that doesn't have the cache and another one that does:
- https://github.com/stackbuilders/dotenv-hs/runs/2737105474?check_suite_focus=true
- https://github.com/stackbuilders/dotenv-hs/runs/2737696708?check_suite_focus=true

However, now that you mention it, I don't see significant improvements in the running time of the one that used the cache. I'm not sure if I missed something, I don't know much about GH actions, maybe you have an idea (?)

CristhianMotoche (approved):
Hey @fefi95 LGTM! Good job! Feel free to merge it once the CI passes and you consider it ready.

fefi95:
Hi @sestrella, I'm putting you in the loop in case you have some input about this. Let me know what you think, Thanks!

sestrella (approved):
@fefi95 overall looks good to me, nice job! I think we just need to update the status badge

sestrella:
@CristhianMotoche BTW I noticed we are still supporting some old GHC versions, perhaps is something we should take a look at later?

fefi95:
Thanks for taking a look @sestrella, I updated the status badge 🙂  Could you remove Travis CI from the checks to allow merging? Not sure if you have access to those settings but it's waiting for that to enable the merge button.

Also, I updated the script to fix the issue with the cache that @CristhianMotoche and I were discussing. Using the `v2-` versions of the commands made it work but I had to add `hspec-discover` in the [build-tools for the tests](https://github.com/stackbuilders/dotenv-hs/pull/129/files#diff-cc1571a6fed7687a75c85d17c8fcb53864155fc7021047ec4bed04d2ff9a025dR159), with that I could also update to cabal 3.4 instead of 2.4.

Here are the 2 runs: when [there's no cache saved](https://github.com/stackbuilders/dotenv-hs/actions/runs/915383554) and when [there is](https://github.com/stackbuilders/dotenv-hs/actions/runs/915453833)

sestrella (approved):
@fefi95 LGTM

PR #130, opened 2021-06-08
Release to Github workflow

fefi95:
# Issue: #128 

##  Description
This PR sets up GH actions to create a release on Github and upload the assets for Linux and MacOS versions of the library.

### DIFF
--- a/.github/workflows/release-gh.yml
+++ b/.github/workflows/release-gh.yml
@@ -1,4 +1,4 @@
-name: Release to package to Github
+name: Release package to Github
 
 on:
   push:
--- a/.github/workflows/release-gh.yml
+++ b/.github/workflows/release-gh.yml
@@ -1,0 +1,41 @@
+name: Release to package to Github
+
+on:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-gh.yml
```diff
@@ -0,0 +1,41 @@
+name: Release to package to Github
```
CristhianMotoche, line 1: typo?
```suggestion
name: Release package to Github
```

CristhianMotoche (approved):
Hey @fefi95 Looks pretty good. Could you please merge it and then test it with some tag (e.g. `v.test`? We can remove that tag and release later.

.github/workflows/release-gh.yml
fefi95: Yes! Thanks for catching that 😄

fefi95:
@CristhianMotoche Thanks for taking a look! Yes, I'll test it with that tag.

Issue #133, opened 2021-06-17
Drop support for GHC 7.10.3


fefi95:
`GHC 7.10.3` is more than 5 years old so could drop the support for this version, this would remove a CI job, potentially some code via CPP and update some upper bounds for the constraints.

CristhianMotoche:
Solved in #135 Thanks @fefi95

PR #135, opened 2021-06-17
Drop support GHC 7.10

fefi95:
# Issue #133 

## Description

- Removes the 7.10.3 job from the CI.
- Updates the base version.
- Removes CPP code that is no longer needed
- Updates the changelog and the version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
 ## MASTER
+
+## Dotenv 0.8.1.0
+* Correct bounds for base. GHC support for versions older than 8.0 was dropped.
+
 ## Dotenv 0.8.0.7
 * Allow megaparsec-0.9.0.0
 
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -4,7 +4,7 @@
 module Main where
 
 import Data.Version (showVersion)
-#if MIN_VERSION_optparse_applicative(0,13,0)
+#if !MIN_VERSION_base(4,13,0)
 import Data.Monoid ((<>))
 #endif
 
--- a/spec/Configuration/DotenvSpec.hs
+++ b/spec/Configuration/DotenvSpec.hs
@@ -26,13 +26,6 @@ import Control.Monad (liftM, void)
 import Data.Maybe (fromMaybe)
 import qualified Data.Text as T
 import qualified Data.Map.Lazy as M
-#if !MIN_VERSION_base(4,8,0)
-import Data.Functor ((<$>))
-#endif
-
-#if !MIN_VERSION_base(4,8,0)
-import Control.Applicative ((<$))
-#endif
 
 main :: IO ()
 main = hspec spec
--- a/src/Configuration/Dotenv/Environment.hs
+++ b/src/Configuration/Dotenv/Environment.hs
@@ -8,20 +8,10 @@ module Configuration.Dotenv.Environment
   ) where
 
 #if MIN_VERSION_base(4,11,0)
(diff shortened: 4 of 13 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 ## MASTER
+
+## Dotenv 0.8.0.7
```
CristhianMotoche, line 3:
```suggestion
## Dotenv 0.8.1.0
```

app/Main.hs
```diff
@@ -4,7 +4,7 @@
 module Main where
 
 import Data.Version (showVersion)
-#if MIN_VERSION_optparse_applicative(0,13,0)
+#if !MIN_VERSION_base(4,13,0)
```
CristhianMotoche, line 7: @fefi95 Why do we need to keep this? Is it because of GHC 8.0? If so, we might want to get rid of that version, too.

spec/Configuration/DotenvSpec.hs
```diff
@@ -26,13 +26,6 @@ import Control.Monad (liftM, void)
 import Data.Maybe (fromMaybe)
 import qualified Data.Text as T
 import qualified Data.Map.Lazy as M
-#if !MIN_VERSION_base(4,8,0)
-import Data.Functor ((<$>))
-#endif
-
-#if !MIN_VERSION_base(4,8,0)
-import Control.Applicative ((<$))
-#endif
```
CristhianMotoche, line 35: :rocket:

src/Configuration/Dotenv/Environment.hs
```diff
@@ -8,20 +8,10 @@ module Configuration.Dotenv.Environment
...
-import System.Environment (getEnvironment, setEnv, unsetEnv)
-#else
-import System.Environment.Compat (getEnvironment, setEnv, unsetEnv)
-#endif
-#endif
-
-#if MIN_VERSION_base(4,11,0)
-import System.Environment.Blank (getEnv)
-#else
-import System.Environment (lookupEnv)
```
CristhianMotoche, line 24: :muscle:

CristhianMotoche (commented):
@fefi95 Overall LGTM! I just left a couple of suggestions. Thanks for the contribution.

app/Main.hs
fefi95: That version and GHC 8.2, if I remember correctly. I tried to remove it and the CI failed https://github.com/stackbuilders/dotenv-hs/actions/runs/947051275. 

Do you want me to remove support for GHC 8.0 as well?

CHANGELOG.md
fefi95: Nice catch! Thank!

app/Main.hs
CristhianMotoche: Snoyberg [suggest to support only the last three versions of GHC](https://www.snoyman.com/blog/2018/07/stop-supporting-older-ghcs/) which means 9.0, 8.10, and 8.8. Nevertheless, that's just a rule of thumb is not a strict rule. Therefore, I think it's fine if we support 8.0, 8.2, 8.4 and 8.6 for now. I'd strongly recommend to stop supporting 8.0, and 8.2 at the beginning of next year (Jan 2022). In conclusion, let's not drop support for GHC 8.0 and 8.2, yet.

CristhianMotoche (approved):
@fefi95 Looks good to me!

app/Main.hs
fefi95: Sure, makes sense! Thanks.

PR #142, opened 2022-06-24
Parse multi-word command interpolations

pbrisbin:
The previous implementation only parsed alphanumeric values within
`$()`, which meant reasonable commands like `foo.sh`, `bin/foo`, or
`foo-bar` would all be rejected.

This commit does more than relax the character set, it uses the
[shellwords][] package to parse multiple words, as a shell would, then
passes the command and any arguments to `proc`.

[shellwords]: https://hackage.haskell.org/package/shellwords

This ensures correct parsing of strings like,

    echo hi\ there
    => ["echo", "hi there"]

    cmd --"this is valid"
    => ["cmd", "--this is valid"]

    cmd --this="Is \"also\" valid"
    => ["cmd", "--this=Is \"also\" valid"]

A note about security: this implementation was originally suggested as a
safer alternative to slurping up all content between `$(` and `)` and
passing it to `shell`. I now see that it is not actually any safer, as
the following example should show:

    FOO=$(echo 'if this were unsafe to pass to shell')
    BAR=$(sh -c "echo 'then this is just as unsafe'")

If we truly don't trust this input, the whole idea of interpolated
commands is suspect. The original mechanism is problematic even with the
restricted parser, and a naive parse-to-shell would be no more or less
problematic.

I think it's reasonable to trust the input, which means any of these
approaches are acceptable. In an ideal world, we'd add a switch in
`Config` to disable this feature for the security concious.

That said, shellwords does bring (at least) two conveniences:

- It correctly handles escaped or quoted ")" characters

  For example,

      FILES=$( find \( -name '*.hs' -o -name '*.hi' \) )

  Without shellwords, we'd still want to account for this ourselves, so
  I think a totally naive parser is off the table.

- It sets up the process tree without an intermediate shell

  This almost certainly doesn't matter as these are so short-lived, but
  in rare cases it can ensure better signal handling.

### DIFF
--- a/spec/Configuration/Dotenv/ParseSpec.hs
+++ b/spec/Configuration/Dotenv/ParseSpec.hs
@@ -7,7 +7,7 @@ import Configuration.Dotenv.ParsedVariable (ParsedVariable(..),
                                             VarValue(..),
                                             VarFragment(..))
 import Data.Void (Void)
-import Test.Hspec (it, describe, Spec, hspec)
+import Test.Hspec (it, context, describe, Spec, hspec)
 import Test.Hspec.Megaparsec (shouldParse, shouldFailOn, shouldSucceedOn)
 import Text.Megaparsec (ParseErrorBundle, parse)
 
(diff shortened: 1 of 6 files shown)


spec/Configuration/Dotenv/ParseSpec.hs
```diff
@@ -142,11 +142,26 @@ spec = describe "parse" $ do
...
+    it "parses a command with arguments" $ do
+      parseConfig "FOO=$(command arg1 arg2)"
+        `shouldParse` [ParsedVariable "FOO" (Unquoted [CommandInterpolation "command" ["arg1", "arg2"]])]
+
+    it "parses a command with quoted arguments" $ do
+      parseConfig "FOO=$(command \"arg 1\" arg2)"
+        `shouldParse` [ParsedVariable "FOO" (Unquoted [CommandInterpolation "command" ["arg 1", "arg2"]])]
+
+    it "parses a command with arguments separated by newlines" $ do
+      parseConfig "FOO=$(command \"arg 1\"\narg2\n)"
```
CristhianMotoche, line 163: @pbrisbin Could you please add a couple of tests to cover the cases that you mention in the issue:
1. `parseConfig "FOO=$(foo-bar.sh)"`
2. `parseConfig "FOO=$(foo/bar.sh)"`

?
Thanks!

src/Configuration/Dotenv/Parse.hs
```diff
@@ -76,9 +76,11 @@ interpolatedValueVarInterpolation = VarInterpolation <$>
...
 
 interpolatedValueCommandInterpolation :: Parser VarFragment
-interpolatedValueCommandInterpolation =
-  CommandInterpolation
-    <$> between (symbol "$(") (symbol ")") (many alphaNumChar)
+interpolatedValueCommandInterpolation = do
+  ws <- between (symbol "$(") (symbol ")") ShellWords.parser
+  pure $ case ws of
+      (commandName:arguments) -> CommandInterpolation commandName arguments
+      _ -> CommandInterpolation "echo" [""]
```
CristhianMotoche, line 83: @pbrisbin Does this mean that if there are no arguments it will return the result of "echo ''"? What about using `VarLiteral ""` instead? It will cause the same effect, won't it?

CristhianMotoche (commented):
@pbrisbin LGTM! I just left a couple of questions. Let me know what you think.

spec/Configuration/Dotenv/ParseSpec.hs
pbrisbin: Sure, will do!

src/Configuration/Dotenv/Parse.hs
pbrisbin: That's right. `VarLiteral ""` is indeed a simpler way to do that :man_facepalming: 

I meant to comment and ask you what you wanted `FOO=$()` to mean. If an empty value works for you, I'll make the change to `VarLiteral`.

stack.yaml
```diff
@@ -3,3 +3,4 @@ packages:
 - '.'
 extra-deps:
 - optparse-applicative-0.17.0.0
+- shellwords-0.1.3.0
```
pbrisbin, line 6: Note that I just added `shellwords` to Stackage, so it'll show up in an LTS soon.

CristhianMotoche (approved):

pbrisbin:
Hi there- is it possible to get a release including this feature soon?

CristhianMotoche:
Hi @pbrisbin My bad. I'll upload a newer version right away.

Issue #146, opened 2022-09-27
Verbose option (for CLI and command)


CristhianMotoche:
It could be useful to get some insights from `dotenv` before loading the environment variables. At the moment, it doesn't

FranzGB:
I would like to work on this issue. The feature will log out the suggestions proposed in this issue:
- log loaded variables on console
- warn on duplicated files 
- log on overriding env variables
but l will let you know if I come up with anything else that could be useful.

I have a question on regards overriding log it means that it checks setted env variables keys for duplicates against the loaded env file right?

CristhianMotoche:
Hey @FranzGB Let's go little by little. I think it was too broad when I opened the PR. Let's focus on:

* Add an options (to the `Config` data type and the CLI flags) to enable the verbose mode
* If the verbose mode is enabled, I want to get a list of the environment variables listed as they're loaded
* If the verbose mode is disabled, I'll just run the command without logging

Once that is done, we can create new tickets for the other points:
* warn on duplicated files (which can be part of #147)
* info about overriding env variables 

Thanks!

FranzGB:
Sounds good to me! I am also halfway to those other points I would like to work on those as well

CristhianMotoche:
Fixed on #148

PR #148, opened 2022-09-29
Feat: Add Verbose option

FranzGB:
This PR adds the ```--verbose``` flag  which enables the CLI to: 
- Log when variables are loaded
- Adds Monad Reader

This implements issue #146

### DIFF
--- a/src/Configuration/Dotenv.hs
+++ b/src/Configuration/Dotenv.hs
@@ -17,7 +17,6 @@ module Configuration.Dotenv
     , loadFile
     , parseFile
     , onMissingFile
-    , DotEnv
       -- * Dotenv Types
     , module Configuration.Dotenv.Types) where
 
@@ -41,10 +40,12 @@ type DotEnv m a = ReaderT Config m a
 
 -- | Loads the given list of options into the environment. Optionally
 -- override existing variables with values from Dotenv files.
-load :: MonadIO m
-     => [(String, String)] -- ^ List of values to be set in environment
-     -> DotEnv m ()
-load = mapM_ applySetting
+load ::
+  MonadIO m =>
+  Bool -- ^ Override existing settings?
+  -> [(String, String)] -- ^ List of values to be set in environment
+  -> m ()
+load override kv = runReaderT (mapM_ applySetting kv) defaultConfig { configOverride = override }
 
 -- | @loadFile@ parses the environment variables defined in the dotenv example
 -- file and checks if they are defined in the dotenv file or in the environment.
--- a/src/Configuration/Dotenv.hs
+++ b/src/Configuration/Dotenv.hs
@@ -52,10 +52,10 @@ load override kv = runReaderT (mapM_ applySetting kv) defaultConfig { configOver
 -- It also allows to override the environment variables defined in the environment
 -- with the values defined in the dotenv file.
 loadFile
-  :: MonadIO m =>
-  Config ->
-  m [(String, String)]
-loadFile = runReaderT (do
+  :: MonadIO m
+  => Config -- ^ Dotenv configuration
+  -> m [(String, String)] -- ^ Environment variables loaded
+loadFile = runReaderT $ do
   Config { .. } <- ask
   environment <- (liftReaderT . liftIO) getEnvironment
   readedVars <- liftReaderT $ fmap concat (mapM parseFile configPath)
@@ -72,7 +72,7 @@ loadFile = runReaderT (do
                     $ "Missing env vars! Please, check (this/these) var(s) (is/are) set:"
                     ++ concatMap ((++) " " . fst) neededVars
              else readedVars
-  mapM applySetting vars)
+  mapM applySetting vars
 
 -- | Parses the given dotenv file and returns values /without/ adding them to
 -- the environment.
--- a/src/Configuration/Dotenv.hs
+++ b/src/Configuration/Dotenv.hs
@@ -44,7 +44,7 @@ load ::
   Bool -- ^ Override existing settings?
   -> [(String, String)] -- ^ List of values to be set in environment
   -> m ()
-load override = mapM_ (applySetting override)
+load override = mapM_ (applySetting override False)
 
 -- | @loadFile@ parses the environment variables defined in the dotenv example
 -- file and checks if they are defined in the dotenv file or in the environment.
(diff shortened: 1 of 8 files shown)


src/Configuration/Dotenv.hs
```diff
@@ -12,95 +12,107 @@
...
 -- override existing variables with values from Dotenv files.
-load ::
-  MonadIO m =>
-  Bool -- ^ Override existing settings?
-  -> [(String, String)] -- ^ List of values to be set in environment
-  -> m ()
-load override = mapM_ (applySetting override)
+load :: MonadIO m
+     => [(String, String)] -- ^ List of values to be set in environment
+     -> DotEnv m ()
```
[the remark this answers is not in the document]
FranzGB: I can change this back 🙌  thanks for the feedback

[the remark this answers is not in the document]
FranzGB: Do we still wait for the boolean value to be passed to this function? In that case, I suggest doing something like this:
```load ::
  MonadIO m =>
  Bool -- ^ Override existing settings?
  -> [(String, String)] -- ^ List of values to be set in environment
  -> m ()
load override kv = runReaderT (mapM_ applySetting kv) defaultConfig { configOverride = override }
```
I think this function is used in the testing suite. WDYT?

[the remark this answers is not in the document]
CristhianMotoche: Yes, let's keep the same signature as before. That means, passing the override variable as a paramater.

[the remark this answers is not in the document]
FranzGB: Alright! Done

src/Configuration/Dotenv.hs
```diff
@@ -12,95 +12,108 @@
...
 loadFile
-  :: MonadIO m
-  => Config -- ^ Dotenv configuration
-  -> m [(String, String)] -- ^ Environment variables loaded
-loadFile Config{..} = do
-  environment <- liftIO getEnvironment
-  readedVars <- concat `liftM` mapM parseFile configPath
-  neededVars <- concat `liftM` mapM parseFile configExamplePath
-  let coincidences = (environment `union` readedVars) `intersectEnvs` neededVars
+  :: MonadIO m => DotEnv m [(String, String)]
```
CristhianMotoche, line 55: @FranzGB Let's not change the signature of any of the exported functions. We can keep the previous signature by doing this:
```hs
loadFile config = runReader config $ do
  Config { .. } <- ask
  ... 
```

As we discussed internally, I want to hide the details of the `Dotenv` monad from the outside world because it's just an implementation detail: 

[Encapsulation](https://en.wikipedia.org/wiki/Encapsulation_(computer_programming)) :wink:

CristhianMotoche: Also, I don't want to change the user interface just for an implementation detail.

src/Configuration/Dotenv.hs
```diff
@@ -12,95 +12,108 @@
...
+
+-- | The function logs in console when a variable is loaded into the
+-- environment.
+info :: MonadIO m => (String, String) -> DotEnv m ()
+info (key, value) = do
+  Config { .. } <- ask
+  when configVerbose
+    $ liftReaderT . liftIO
+    $ putStrLn
+    $ "[INFO]: Load env '" ++ key ++ "' with value '" ++ value ++ "'"
```
CristhianMotoche, line 109: This looks nice!

src/Configuration/Dotenv.hs
```diff
@@ -54,53 +55,67 @@ loadFile
...
-loadFile Config{..} = do
-  environment <- liftIO getEnvironment
-  readedVars <- concat `liftM` mapM parseFile configPath
-  neededVars <- concat `liftM` mapM parseFile configExamplePath
-  let coincidences = (environment `union` readedVars) `intersectEnvs` neededVars
+loadFile = runReaderT $ do
+  Config { .. } <- ask
+  environment <- (liftReaderT . liftIO) getEnvironment
+  readedVars <- liftReaderT $ fmap concat (mapM parseFile configPath)
+  neededVars <- liftReaderT $ fmap concat (mapM parseFile configExamplePath)
```
CristhianMotoche, line 62: Hey @FranzGB We can save the lifting by leaving the code as it was before and call `applySetting` with `runReaderT` at the end:

```suggestion
lloadFile config@Config{..} = do
  environment <- liftIO getEnvironment
  readedVars <- concat `liftM` mapM parseFile configPath
  neededVars <- concat `liftM` mapM parseFile configExamplePath
  ...
  runReaderT config (mapM applySetting vars)
```
Would that work?

src/Configuration/Dotenv/Types.hs
```diff
@@ -30,4 +36,29 @@ defaultConfig =
     { configExamplePath = []
     , configOverride = False
     , configPath = [ ".env" ]
+    , configVerbose = False
     }
+
+
+newtype ReaderT r m a = ReaderT { runReaderT :: r -> m a }
+mapReaderT :: (m a -> n b) -> ReaderT r m a -> ReaderT r n b
+mapReaderT f m = ReaderT $ f . runReaderT m
```
CristhianMotoche, line 45: @FranzGB Are we using this function? If not, can we remove it?

src/Configuration/Dotenv/Types.hs
```diff
@@ -30,4 +36,29 @@ defaultConfig =
...
+instance (Monad m) => Monad (ReaderT r m) where
+    return   = liftReaderT . return
+    m >>= k  = ReaderT $ \ r -> do
+      a <- runReaderT m r
+      runReaderT (k a) r
+    m >> k = ReaderT $ \ r -> runReaderT m r >> runReaderT k r
+
+liftReaderT :: m a -> ReaderT r m a
+liftReaderT m = ReaderT (const m)
+{-# INLINE liftReaderT #-}
```
CristhianMotoche, line 61: @FranzGB Why `INLINE`?

src/Configuration/Dotenv.hs
FranzGB: I think that could work I will change this in a bit 👍

src/Configuration/Dotenv/Types.hs
FranzGB: We are using this function on the instance `Functor fmap` of the ReaderT

src/Configuration/Dotenv/Types.hs
FranzGB: Woops I copied this function from a library and forgot to remove that

src/Configuration/Dotenv/Types.hs
FranzGB: I remove the export should I remove then instance as well?

CristhianMotoche: Let's not remove the instance. Thanks!

CristhianMotoche (approved):
:shipit: This looks like a good start. Thanks @FranzGB !

Issue #147, opened 2022-09-27
Allow/Forbid duplicated env variables in file


CristhianMotoche:
At the moment, the dotenv file load the first value from the file (refer to issue #121) which may/may not be expected. I would like to change this to allow/forbid duplicated env variables in the `.env` files. Something like this:
```
cat > .env

dotenv 'echo $FOO' --no-dups
[ERROR]: Env 'FOO' is duplicated in a dotenv file. Please, fix that (or remove --no-dups).
```

Additionally, we should add a comment in the README to let users know about the value that we'll get at the end if `--no-dups` is not given, which is the first env variable in the files at the moment.

FranzGB:
I would like to work on this. What would be the default behavior? also should dotenv raise an error instead of continuing with the process?

CristhianMotoche:
Hey @FranzGB To answer your questions:

> What would be the default behavior?

The default behaviour should be to allow duplicated variables and specifying in the README that we take the last one found to set up.

> also should dotenv raise an error instead of continuing with the process?

Yes, please, let's display an error like the one I mentioned above:

> $ dotenv 'echo $FOO' --no-dups
> [ERROR]: Env 'FOO' is duplicated in a dotenv file. Please, fix that (or remove --no-dups).

Thanks for contributing to Dotenv!

PR #149, opened 2022-10-14
feat: No duplicates option 

FranzGB:
This PR implements the solution for #147

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -34,7 +34,7 @@ main = do
           { configExamplePath = dotenvExampleFiles
           , configOverride = override
           , configVerbose = verbose
-          , configDuplicates = duplicates
+          , disallowDuplicates = duplicates
           , configPath =
               if null dotenvFiles
                 then configPath defaultConfig
@@ -72,6 +72,11 @@ config = Options
 
      <*> switch (  long "verbose"
                   <> help "Specify this flag to print out the variables loaded and other useful insights" )
+
+     <*> switch (  long "no-dups"
+                  <> short 'D'
+                  <> help "Specify this flag to disallow duplicate variables"
+                  )
 
      <*> argument str (metavar "PROGRAM")
 
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -34,7 +34,7 @@ main = do
           { configExamplePath = dotenvExampleFiles
           , configOverride = override
           , configVerbose = verbose
-          , disallowDuplicates = duplicates
+          , allowDuplicates = duplicates
           , configPath =
               if null dotenvFiles
                 then configPath defaultConfig
@@ -73,7 +73,7 @@ config = Options
      <*> switch (  long "verbose"
                   <> help "Specify this flag to print out the variables loaded and other useful insights" )
 
-     <*> switch (  long "no-dups"
+     <*> flag True False (  long "no-dups"
                   <> short 'D'
                   <> help "Specify this flag to disallow duplicate variables"
                   )
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -75,7 +75,7 @@ config = Options
 
      <*> flag True False (  long "no-dups"
                   <> short 'D'
-                  <> help "Specify this flag to disallow duplicate variables"
+                  <> help "Specify this flag to disallow duplicate variables and throw an error" 
                   )
 
      <*> argument str (metavar "PROGRAM")
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -75,7 +75,7 @@ config = Options
 
      <*> flag True False (  long "no-dups"
                   <> short 'D'
-                  <> help "Specify this flag to disallow duplicate variables and throw an error" 
+                  <> help "Specify this flag to forbid duplicate variables"
                   )
 
      <*> argument str (metavar "PROGRAM")
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -11,8 +11,8 @@ import           Data.Monoid          ((<>))
 import           Options.Applicative
 import           Paths_dotenv         (version)
 
-import           Control.Monad        (void)
 import           Configuration.Dotenv (Config (..), defaultConfig, loadFile)
+import           Control.Monad        (void)
 import           System.Exit          (exitWith)
 import           System.Process       (system)
 
@@ -21,6 +21,7 @@ data Options = Options
   , dotenvExampleFiles :: [String]
   , override           :: Bool
   , verbose            :: Bool
+  , duplicates         :: Bool
   , program            :: String
   , args               :: [String]
   } deriving (Show)
@@ -33,6 +34,7 @@ main = do
           { configExamplePath = dotenvExampleFiles
           , configOverride = override
           , configVerbose = verbose
+          , configDuplicates = duplicates
           , configPath =
               if null dotenvFiles
                 then configPath defaultConfig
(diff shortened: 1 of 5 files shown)


spec/Configuration/DotenvSpec.hs
```diff
@@ -47,26 +47,26 @@ spec = do
     it "loads the configuration options to the environment from a file" $ do
       lookupEnv "DOTENV" `shouldReturn` Nothing
 
-      void $ loadFile $ Config ["spec/fixtures/.dotenv"] [] False False
+      void $ loadFile $ Config ["spec/fixtures/.dotenv"] [] False False True
```
CristhianMotoche, line 50: @FranzGB What about creating a sample config and change it accordingly:
```hs
sampleConfig = onfig ["spec/fixtures/.dotenv"] [] False False True
-- Then, you can change the properties as needed:
loadFile $ sampleConfig { allowDuplicates = False }
loadFile $ sampleConfig { files = [] }
...
```
? I think that will make things easier to read in the test suite.

spec/Configuration/DotenvSpec.hs
```diff
@@ -125,14 +125,21 @@ spec = do
...
         lookupEnv "DOTENV" `shouldReturn` Just "true"
 
     context "when target file is missing" $
       it "executes supplied handler instead" $
-        onMissingFile (True <$ loadFile (Config ["spec/fixtures/foo"] [] True False)) (return False)
+        onMissingFile (True <$ loadFile (Config ["spec/fixtures/foo"] [] True False True)) (return False)
           `shouldReturn` False
 
+  describe "onDuplicatedKeys" $ after_ clearEnvs $ do
+    context "when target file has duplicated key and the duplicate are not allowed" $
```
CristhianMotoche, line 137:
```suggestion
    context "when target file has duplicated key and the duplicate are forbidden" $
```

src/Configuration/Dotenv.hs
```diff
@@ -89,13 +91,16 @@ applySetting :: MonadIO m =>
...
     then info kv >> setEnv'
     else do
       res <- liftReaderT . liftIO $ lookupEnv k
       case res of
         Nothing -> info kv >> setEnv'
-        Just _  -> return kv
+        Just _  -> onDisallowDuplicates k >>
+                if configOverride
+                  then info kv >> setEnv'
+                  else return kv
```
CristhianMotoche, line 103: Hey @FranzGB Correct me if I'm wrong but I see your proposition here is to throw an error if the `--no-dups` flag is given and the variable is set twice. Am I write? I suppose we won't get an error if the environment variable is already in the system (thanks to the `configOverride`). However, I found this implementation kind of complicated. I would suggest to reduce the [cyclomatic complexity](https://en.m.wikipedia.org/wiki/Cyclomatic_complexity) by double checking duplicated environment variables one step above (before calling `applySetting` on any variable). WDYT?

app/Main.hs
```diff
@@ -71,6 +73,11 @@ config = Options
      <*> switch (  long "verbose"
                   <> help "Specify this flag to print out the variables loaded and other useful insights" )
 
+     <*> flag True False (  long "no-dups"
+                  <> short 'D'
+                  <> help "Specify this flag to disallow duplicate variables and throw an error"
```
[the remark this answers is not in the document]
FranzGB: That is better worded thank you Cris

spec/Configuration/DotenvSpec.hs
FranzGB: Yes I agree that is way better thank you

spec/Configuration/DotenvSpec.hs
```diff
@@ -125,14 +124,27 @@ spec = do
...
 
+  describe "onDuplicatedKeys" $ after_ clearEnvs $ do
+    context "when target file has duplicated key and the duplicate are forbidden" $
+      it "throws an error" $
+        loadFile sampleConfig { configPath = ["spec/fixtures/.dotenv", "spec/fixtures/.dotenv"], allowDuplicates = False }
+          `shouldThrow` anyIOException
+    context "when target file does not have duplicated key and the duplicate are forbidden" $
+      it "works as usual" $ do
+        void $ loadFile sampleConfig { allowDuplicates = False }
+        lookupEnv "DOTENV" `shouldReturn` Just "true"
```
CristhianMotoche, line 143: @FranzGB I think we should double check that the first env variable is set in this case.

src/Configuration/Dotenv.hs
```diff
@@ -8,17 +8,17 @@
...
 {-# LANGUAGE RecordWildCards #-}
 
 module Configuration.Dotenv
-    ( -- * Dotenv Load Functions
-      load
-    , loadFile
-    , parseFile
-    , onMissingFile
+      -- * Dotenv Load Functions
+  ( load
```
CristhianMotoche, line 15: @FranzGB I think the comment should go inside the parenthesis. Could you please double check the haddock documentation?

src/Configuration/Dotenv.hs
```diff
@@ -103,19 +109,35 @@ applySetting kv@(k, v) = do
...
+     MonadCatch m
   => m a -- ^ Action to perform that may fail because of missing file
-  -> m a               -- ^ Action to perform if file is indeed missing
+  -> m a -- ^ Action to perform if file is indeed missing
   -> m a
 onMissingFile f h = catchIf isDoesNotExistError f (const h)
+
+-- | The helper throws an exception if the allow duplicate is set to False.
+onForbidDuplicates :: MonadIO m => String -> m ()
+onForbidDuplicates key =
```
CristhianMotoche, line 130: @FranzGB The name suggests that this is a callback. However, I see it as an standard function. Let's rename it to `forbidDuplicates`. Does that sound good?

src/Configuration/Dotenv.hs
```diff
@@ -103,19 +109,35 @@ applySetting kv@(k, v) = do
...
+  userError $
+  "[ERROR]: Env '" ++
+  key ++
+  "' is duplicated in a dotenv file. Please, fix that (or remove --no-dups)."
+
+lookUpDuplicates :: MonadIO m => [String] -> m ()
+lookUpDuplicates [] = return ()
+lookUpDuplicates [_] = return ()
+lookUpDuplicates (x:xs) =
+  if x `elem` xs
```
CristhianMotoche, line 141: @FranzGB I think this will eventually call `onForbidDuplicates` on any env var when `--no-dups` is given. I see you're checking if the env is in the list of env vars. If so, then call `onForbidDuplicates` and that will throw an error. Then, my guess is that even normal files, it will throw an error. Am I wrong? :thinking:

CristhianMotoche (commented):
Hey @FranzGB I think this looks better. Although, I'm not sure if we're solving the issue without inserting another one. Could you please take a look at the comments that I left?

src/Configuration/Dotenv.hs
FranzGB: For sure

src/Configuration/Dotenv.hs
FranzGB: Sure I will take a look at it

spec/Configuration/DotenvSpec.hs
FranzGB: That's exactly what this test case is doing 🙌

src/Configuration/Dotenv.hs
CristhianMotoche: You're right @FranzGB Thanks! I got all confused.

CristhianMotoche (approved):
:shipit:

PR #152, opened 2022-12-09
chore: Go back to previous return type for loadFile

CristhianMotoche:
# Change
After reviewing the question described by @nsaunders I think it would be best
to return to the initial return value for `loadFile`. It would make more sense
since it's an action that doesn't need to return the list of environment variables
loaded. It's main purpose is to load the envs from a file before running another action.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## MASTER
 ## Dotenv 0.10.0.0
-### Added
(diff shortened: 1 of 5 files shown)


sestrella (approved):
@CristhianMotoche LGTM!

nsaunders:
Glad to see this change, thanks again!

PR #153, opened 2022-12-15
chore: Update CI and fix static upload

CristhianMotoche:
# Changes
- Compile the project with `ghc-musl` to generate the static binary
- Compress binary file with `upx` (from 26 MB to 2MB!!!)
- Upload binary file to GH draft
- New workflow to release to Hackage on version published

I followed a different approach from `hapistrano` since we don't need to generate a `dotenv` docker image.

Build sample: [here](https://github.com/stackbuilders/dotenv-hs/actions/runs/3722568027/jobs/6313500421)
Draft release: [here](https://github.com/stackbuilders/dotenv-hs/releases/tag/untagged-bb61dfab34446a3fd3ef)


CristhianMotoche:
Thanks @sestrella for the tips on the Hackage automation. Please, give it a look to this PR an let me know if it looks good. Thanks in advance!

.github/workflows/publish.yml
```diff
@@ -0,0 +1,54 @@
...
+
+      - name: Upload Docs for candidateto Hackage
+        run: |
+          TAG_VERSION=${{ github.ref_name }}
+          VERSION=${TAG_VERSION:1}
+          DOCS_PATH=$(ls dist-newstyle/dotenv-*-docs.tar.gz)
+          curl -X PUT \
+           -H 'Content-Type: application/x-tar' \
+           -H 'Content-Encoding: gzip' \
+           -H 'Authorization: X-ApiKey ${{ secrets.HACKAGE_API_KEY }}' \
```
sestrella, line 52: @CristhianMotoche based on the official documentation, GH automatically redacts secrets:

> Warning: GitHub automatically redacts secrets printed to the log, but you should avoid printing secrets to the log intentionally.

However, most of the examples found in the documentation define an environment variable based on a secret and then reference the variable in the script, here is a code snippet from GH:

```
steps:
  - shell: bash
    env:
      SUPER_SECRET: ${{ secrets.SuperSecret }}
    run: |
      example-command "$SUPER_SECRET"
```

Therefore, I would recommend following the same approach here. What do you think?

https://docs.github.com/en/actions/security-guides/encrypted-secrets#example-using-bash

CristhianMotoche: Good catch @sestrella Thanks for the tip!

CristhianMotoche: Hey @sestrella Regardless the GH documentation, I see the secrets are actually hidden somehow. Take a look at this sample: https://github.com/stackbuilders/dotenv-hs/actions/runs/3735107653/jobs/6337959409#step:7:10 I'm going to implement  your suggestion anyway, though. To avoid any possible vulnerability.

sestrella: Right, GH automatically redacts secrets, however, sensitive data is usually passed as environment variables to bash scripts, so I think that is why the documentation recommends that approach. Thank you for taking care of this!

sestrella (approved):
@CristhianMotoche amazing job! LGTM

PR #161, opened 2023-04-13
Introduced nubByLastVar vars before applying setting.

anddriex:
- This PR addresses [issue #121](https://github.com/stackbuilders/dotenv-hs/issues/121)
- Mapping vars before setting in environment using function `nubByLastVar`.
- Added function [`nubByLastVar`](https://github.com/stackbuilders/dotenv-hs/compare/issue-121-dotenv-picks-last-value?expand=1#diff-28752b99ee8c901d2f3dead57a3943b5fe22dc7c15bf7ec272eee5e646f4843fR163) that yields a [`Data.Map`](https://hackage.haskell.org/package/containers-0.4.0.0/docs/Data-Map.html)  from a list and the returns back an ordered list of type `[(String), (String)]`. If the list contains more than one value for the same key, the last value for the key is retained.

### DIFF
--- a/spec/fixtures/.dotenv.dup
+++ b/spec/fixtures/.dotenv.dup
@@ -1,0 +1,3 @@
+FOO=first
+BAR=tender
+FOO=last
--- a/src/Configuration/Dotenv.hs
+++ b/src/Configuration/Dotenv.hs
@@ -24,17 +24,19 @@ import           Configuration.Dotenv.Environment    (getEnvironment, lookupEnv,
                                                       setEnv)
 import           Configuration.Dotenv.Parse          (configParser)
 import           Configuration.Dotenv.ParsedVariable (interpolateParsedVariables)
-import           Configuration.Dotenv.Types          (Config (..), defaultConfig)
-import           Control.Monad.Trans                 (lift)
-import           Control.Monad.Reader                (ReaderT, ask, runReaderT)
+import           Configuration.Dotenv.Types          (Config (..),
+                                                      defaultConfig)
 import           Control.Exception                   (throw)
 import           Control.Monad                       (unless, when)
 import           Control.Monad.Catch
 import           Control.Monad.IO.Class              (MonadIO (..))
+import           Control.Monad.Reader                (ReaderT, ask, runReaderT)
+import           Control.Monad.Trans                 (lift)
 import           Data.Function                       (on)
-import           Data.List.NonEmpty                  (NonEmpty(..))
-import qualified Data.List.NonEmpty as NE
-import           Data.List                           ((\\), intercalate, union)
+import           Data.List                           (intercalate, union, (\\))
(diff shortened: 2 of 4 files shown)


src/Configuration/Dotenv.hs
```diff
@@ -156,3 +158,6 @@ lookUpDuplicates (x:xs) =
   if x `elem` xs
     then forbidDuplicates x
     else lookUpDuplicates xs
+
+orderList :: [(String, String)] -> [(String, String)]
+orderList = toList . fromList
```
dbalseiro, line 163: I'm curious... why not just use `sortOn fst`?
ref: https://hackage.haskell.org/package/base-4.18.0.0/docs/Data-List.html#v:sortOn

dbalseiro: Oh I see It's because you want to remove duplicates... but [line 90](https://github.com/stackbuilders/dotenv-hs/pull/161/files#diff-28752b99ee8c901d2f3dead57a3943b5fe22dc7c15bf7ec272eee5e646f4843fR91) tells me that duplicates may be allowed :thinking: I think you're introducing an undesired side effect here... or what am I missing?

anddriex: Well, the idea here is to pick the last set value in the list if it contains duplicates. For that I make use of the `Data.Map` to drop duplicates picking the last value for a repeated key but it has the side effect that also orders the list. It seems that when duplicates were allowed it was using the first value and ignoring the last one. Behavior that this PR aims to change.

I see your point about the name, maybe it can be named  something like `mapVars`?

spec/fixtures/.dotenv.dup
```diff
@@ -0,0 +1,3 @@
+FOO=first
+BAR=tender
+FOO=last
```
CristhianMotoche, line 3: @anddriex You many need to add this file in the cabal file [here](https://github.com/stackbuilders/dotenv-hs/blob/master/dotenv.cabal#L48-L50)

CristhianMotoche (commented):
@anddriex Overall LGTM! It seems pretty simple and similar to what other libraries from other languages are doing. Thanks!

I left a suggestion in a file. Could you please review it?

src/Configuration/Dotenv.hs
CristhianMotoche: Hey @anddriex I would suggest a different name. Something like `nubByLastVar` since it's removing duplicates, and taking the last variable. WDYT?

anddriex: Sounds good. Thanks for your suggest!

spec/fixtures/.dotenv.dup
anddriex: Done.

CristhianMotoche (approved):
LGTM! Thanks for working on this @anddriex :shipit:

PR #163, opened 2023-05-04
feat: Add Default instance for Config

aloussase:
I think this would be nice for people who are already using data-default-class.

### DIFF
--- a/dotenv.cabal
+++ b/dotenv.cabal
@@ -98,6 +98,7 @@ library
                        , text
                        , exceptions >= 0.8 && < 0.11
                        , mtl >= 2.2.2 && < 2.4
+                       , data-default-class >= 0.1.2 && < 0.2
 
   hs-source-dirs:      src
   ghc-options:         -Wall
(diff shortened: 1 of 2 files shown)


dotenv.cabal
```diff
@@ -134,6 +135,7 @@ test-suite dotenv-test
                      , exceptions >= 0.8 && < 0.11
                      , hspec-megaparsec >= 2.0 && < 3.0
                      , mtl
+                     , data-default-class >= 0.1.2 && < 0.2
```
CristhianMotoche, line 138: Is it necessary to add it to the test-suite?

CristhianMotoche (commented):
Hey @aloussase Thanks a lot for your contribution. I like the idea to include this instance. I just left one question in your PR. Could you please take a look at it?

dotenv.cabal
aloussase: I think it is, because the test suite specifies the src directory` in `hs-source-dirs`.

CristhianMotoche: Oh, you're right. I'll fix that in a separated PR. Thanks man!

CristhianMotoche (approved):
LGTM! Thank you @aloussase I'll try to get it to hackage as soon as I can.

dotenv.cabal
aloussase: I thought it was on purpose haha you're welcome :)

PR #164, opened 2023-05-10
chore: Expose new modules and update cabal file

CristhianMotoche:
# Change
- Remove `src` from test-suite
- Remove `other_modules` from test-suite
- Reduce dependencies in test-suite
- Expose internal modules
- Document new exposed types and functions

### DIFF
--- a/dotenv.cabal
+++ b/dotenv.cabal
@@ -1,5 +1,5 @@
 name:                dotenv
-version:             0.12.0.0
+version:             0.11.0.1
 synopsis:            Loads environment variables from dotenv files
 homepage:            https://github.com/stackbuilders/dotenv-hs
 description:
@@ -84,8 +84,8 @@ executable dotenv
 library
   exposed-modules:       Configuration.Dotenv
                        , Configuration.Dotenv.Internal
-  other-modules:         Configuration.Dotenv.Environment
-                       , Configuration.Dotenv.Parse
+                       , Configuration.Dotenv.Environment
+  other-modules:         Configuration.Dotenv.Parse
                        , Configuration.Dotenv.ParsedVariable
                        , Configuration.Dotenv.Text
                        , Configuration.Dotenv.Types
--- a/dotenv.cabal
+++ b/dotenv.cabal
@@ -1,5 +1,5 @@
 name:                dotenv
-version:             0.11.0.0
+version:             0.12.0.0
 synopsis:            Loads environment variables from dotenv files
 homepage:            https://github.com/stackbuilders/dotenv-hs
 description:
--- a/dotenv.cabal
+++ b/dotenv.cabal
@@ -84,9 +84,9 @@ executable dotenv
 library
   exposed-modules:      Configuration.Dotenv
                       , Configuration.Dotenv.Environment
+                      , Configuration.Dotenv.Text
   other-modules:        Configuration.Dotenv.Parse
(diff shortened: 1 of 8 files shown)


dotenv.cabal
```diff
@@ -110,32 +110,19 @@ library
 
 test-suite dotenv-test
   type: exitcode-stdio-1.0
-  hs-source-dirs: spec, src
```
sestrella, line 113: 💯

dotenv.cabal
```diff
@@ -84,10 +84,10 @@ executable dotenv
 library
   exposed-modules:      Configuration.Dotenv
                       , Configuration.Dotenv.Environment
-  other-modules:        Configuration.Dotenv.Parse
-                      , Configuration.Dotenv.ParsedVariable
                       , Configuration.Dotenv.Text
-                      , Configuration.Dotenv.Types
+                      , Configuration.Dotenv.ParsedVariable
```
sestrella, line 88: @CristhianMotoche I would recommend using a `Internal` module that re-exports internal details rather than directly exposing them for a few reasons:

- An internal module makes it clear to consumers that its API is more subject to change than the public API.
- Hiding all internal details behind a single module allows for greater refactoring flexibility; for example, you could change the name of the ParsedVariable or split its content without consumers noticing.

CristhianMotoche: Thanks for the suggestion @sestrella ! I updated this PR. Please, let me know if this new module structure look good. We'll need to do a major version release, though, since I stopped exporting `Configuration.Dotenv.Environment` in favor of `Configuration.Dotenv.Internal`.

src/Configuration/Dotenv/Internal.hs
```diff
@@ -0,0 +1,14 @@
+module Configuration.Dotenv.Internal
+  ( module Configuration.Dotenv.Types
```
sestrella, line 2: @CristhianMotoche looks good, but I'd recommend only exporting the functions and types needed for testing to hide some implementation details; the more implementation details we expose, the greater the risk of introducing a breaking change.

CristhianMotoche: Makes sense to me @sestrella I'll keep the export of `Configuration.Dotenv.Environment` to avoid the major release. Please, take a second look at it. Thanks in advance!

dotenv.cabal
```diff
@@ -1,5 +1,5 @@
 name:                dotenv
-version:             0.11.0.0
+version:             0.11.0.1
```
sestrella, line 2: @CristhianMotoche based on the fact we are moving some modules from the `exposed-modules` section to the `other-modules` section, I think we should publish this change as a major release since it would break some builds if another project imports directly one of the modules previously exposed

CristhianMotoche: Hey @sestrella Mmm I didn't move any module away from `exposed-modules`. My only change was to add `Configuration.Dotenv.Internal` as a new exposed module. For that reason, I think this is a minor change. WDYT?

sestrella: @CristhianMotoche my mistake, I misread the diff, you are correct, it appears that we are only adding a new module to the `exposed-modules` section, in this case, a minor release appears to be the best option.

sestrella (approved):
@CristhianMotoche LGTM

CristhianMotoche:
Thank you so much @sestrella I appreciate your review!

PR #171, opened 2023-06-23
chore: Update support up to GHC 9.6 and bump new version 0.12.0.0

CristhianMotoche:
# Changes
- Increase support up to GHC 9.6

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        ghc: ["9.6", "9.4", "9.2"]
+        ghc: ["8.10", "9.0", "9.2", "9.4", "9.6"]
         os: [ubuntu-latest]
         include:
           - os: macos-latest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        ghc: ["9.2", "9.0", "8.10"]
+        ghc: ["9.6", "9.4", "9.2"]
         os: [ubuntu-latest]
         include:
           - os: macos-latest
@@ -32,7 +32,7 @@ jobs:
         uses: haskell/actions/setup@v2
         with:
           ghc-version: ${{ matrix.ghc }}
-          cabal-version: "3.6"
+          cabal-version: "3.8"
 
       - name: Update dependencies with Hackage
         run: cabal update
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,12 @@
 ## MASTER
+## Dotenv 0.12.0.0
+### Modified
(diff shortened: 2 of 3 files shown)


dotenv.cabal
```diff
@@ -34,6 +34,7 @@ description:
 license:             MIT
 license-file:        LICENSE
 author:              Justin Leitgeb
+tested-with:         ghc ==9.0 ghc ==9.4 ghc ==9.6
```
sestrella, line 37:
```suggestion
tested-with:         ghc ==9.2 ghc ==9.4 ghc ==9.6
```

.github/workflows/build.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        ghc: ["9.2", "9.0", "8.10"]
+        ghc: ["9.6", "9.4", "9.2"]
```
sestrella, line 21: @CristhianMotoche if we are going to exclude `9.0` from the build matrix, I think we should bump the version constraint for `base` accordingly

dotenv.cabal
CristhianMotoche: Good catch!

.github/workflows/build.yml
CristhianMotoche: That's a good observation @sestrella I will preserve the testing against GHC 8.10 since dotenv can be used in  projects with that version of GHC without any inconveniences. Although, I do need to change the base boundaries since [GHC 8.10 is shipped with base 4.14](https://wiki.haskell.org/Base_package).

dotenv.cabal
```diff
@@ -62,7 +63,7 @@ flag dev
 
 executable dotenv
   main-is:             Main.hs
-  build-depends:         base >= 4.9 && < 5.0
+  build-depends:         base >= 4.14 && < 5.0
```
sestrella, line 66: @CristhianMotoche this change appears to be effectively dropping support for GHC versions older than 8.10, I believe we should document this in the CHANGELOG.

image

CHANGELOG.md
```diff
@@ -1,4 +1,12 @@
 ## MASTER
+## Dotenv 0.12.0.0
```
sestrella, line 2: @CristhianMotoche I believe this is beyond the scope of the discussion, but I'd like to add a random thought here:

We previously discussed `semantic-release` which does not support PVP versioning per se, however, I believe we could do something with this or other similar tools to automate the CHANGELOG generation based on the commit history. What do you think?

References:
- https://github.com/semantic-release/release-notes-generator

CristhianMotoche: Hey @sestrella I like the idea and it seems it will help to generate release notes. Although, I think we will need to strictly follow the semantic release commit format, right?

sestrella: @CristhianMotoche yes, standardizing commit messages would be required to make that work, as most of these tools generate a CHANGELOG based on the git history

FranzGB (approved):
Looks good to me. Thanks @CristhianMotoche

PR #175, opened 2023-07-11
feat: Add FromEnv to generically create values from env variables

aloussase:
Hi! I added a small class to be able to create values from environment variables, like this:
```haskell
newtype Config = Config { s3Bucket :: String } deriving (Generic, FromEnv)

main = do
    -- This returns m (Maybe Config)
    -- Where m is some MonadIO
    -- Also, we mostly don't need this type application depending on the context
    config <- fromEnv_ @Config
```
I think this would be a nice addition to this library. I am new to generic programming, so I
would highly appreciate any feedback to make this good enough to be included in the library.

Finally, I added some tests and documented the public api in `Configuration.Dotenv.FromEnv`.

### DIFF
--- a/spec/Configuration/Dotenv/FromEnvSpec.hs
+++ b/spec/Configuration/Dotenv/FromEnvSpec.hs
@@ -1,0 +1,57 @@
+{-# LANGUAGE DataKinds        #-}
+{-# LANGUAGE DeriveAnyClass   #-}
+{-# LANGUAGE DeriveGeneric    #-}
+{-# LANGUAGE TypeApplications #-}
+module Configuration.Dotenv.FromEnvSpec (main, spec) where
+
+import           GHC.Generics
+import           Test.Hspec                       (Spec, after_, describe,
+                                                   hspec, it, shouldBe,
+                                                   shouldSatisfy)
+
+import           Configuration.Dotenv.Environment (getEnvironment, setEnv,
+                                                   unsetEnv)
+import           Configuration.Dotenv.FromEnv
+import           Data.Maybe                       (fromJust, isJust)
+
+data Config = Config
+  { dbURL  :: !String
+  , apiKey :: !String
+  }
+  deriving (Eq, Show, Generic, FromEnv)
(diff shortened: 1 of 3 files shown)


spec/Configuration/Dotenv/FromEnvSpec.hs
```diff
@@ -0,0 +1,57 @@
...
+      config `shouldSatisfy` isJust
+      fromJust config `shouldBe` Config "hello" "world"
+
+    it "returns the configuration object when using different names for the environment variables" $ do
+      setEnv "DB_URL" "hello"
+      setEnv "API_KEY" "world"
+      let converter "dbURL"  = Just "DB_URL"
+          converter "apiKey" = Just "API_KEY"
+          converter _        = Nothing
+      config <- fromEnv converter
```
CristhianMotoche, line 51: I wish there were a way to get the type without the need of the converter. 🤔

aloussase: Yeah, me too. At first I tried doing as optparse-generic and use a type level tag to specify what the field name in the .env file is, but that approach requires a wrapper type and feels like a leaky abstraction. One option would be to define default converters and let the user specify one if needed. I would also like a cleaner way.

aloussase: I added `GFromEnv`. Now users may just derive `FromEnv` for the default behavior, or provide an implementation using `gFromEnv` with custom options. I took a look at what aeson does and implemented something similar.
The downside is that to provide default options I included a dependency on the `casing` library.

CristhianMotoche:
Hey @aloussase The approach is interesting and looks cool. Although, I think this could be over scope of dotenv. What are the possible use cases for this new feature?

aloussase:
Yes I agree it may not be such a good fit for this library. I thought it would be a good idea because if I am loading env variables using this library it would be useful to have a way to construct some kind of configuration object from the loaded environment variables.

PR #188, opened 2024-01-02
fix: Replace haskell/actions with haskell-actions

CristhianMotoche:
# Changes
- `s/haskell\/actions/haskell-actions`


cptrodolfox (approved):
:rocket:

Issue #64, opened 2017-09-12
.envs as Ciphertext


CristhianMotoche:
## Motivation
Setting things like passwords, key phrases, tokens, and so on is very common in .envs. As well as sharing `.envs`, specially when someone new enters into a project. Of course, having a security culture won't allow that this kind of things happen. However, we can make sure that they don't happen by adding `.envs` as cipher text.

## Description
Given a `.env` and a password we can generate a `.env.cipher` that will have the `.env` vars as ciphertext.
![image](https://user-images.githubusercontent.com/8370088/30328557-4ead0e4e-9795-11e7-856b-cc7aeeef2f25.png)

In order to load this `.env` vars, we could set a flag in the command line. Internally it will load the env vars in the environment.
![image](https://user-images.githubusercontent.com/8370088/30328437-f7673d4e-9794-11e7-8f8a-db2c70621b40.png)

It will be similar to the way ansible vault saves secret files.

@jsl @juanpaucar @sestrella 
This is just an idea that came to my mind in a dream (just kidding). I'd like to know what do you think. Is it worth to work on it? Or add it?

po5i:
![](https://github.trello.services/images/mini-trello-icon.png) [.envs as Ciphertext](https://trello.com/c/ax7BXjfi/39-envs-as-ciphertext)

CristhianMotoche:
I think there is already a library for this purpose, thought, the variable codification is different: https://github.com/CollegeVine/confcrypt#readme
In case we want to add this complexity, I think it would be better to have a second repo for this purpose.

PR #190, opened 2024-02-06
Added `kleidukos/get-tested` action to generate CI matrix #64

### LINKED ISSUE #189: Use `kleidukos/get-tested` and include GHC supported versions in cabal file
Let's follow the same approach that Rollbar Haskell is doing: https://github.com/stackbuilders/rollbar-haskell/pull/64

DavidMazarro:
Thanks to the [get-tested](https://github.com/Kleidukos/get-tested) action, we can automatically generate the matrix of GHC versions to run the CI on. This keeps the `tested-with` versions we set in Cabal files accurate to what is actually being tested in the CI.

Since the action takes one Cabal file as reference, I pointed it to `rollbar-client.cabal` since that is the main library in this repository. In any case, us maintainers need to keep all of the `tested-with` versions in sync across the four Haskell packages we offer in this repository.

This PR also changes the `tested-with` versions of the Cabal in the project to the latest minor versions of GHC 8.10 (the latest GHC 8 version), GHC 9.6 and GHC 9.8 (the two most recent GHC 9 versions).

Closes #189.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,6 +26,7 @@ jobs:
         with:
           cabal-file: dotenv.cabal
           ubuntu: true
+          macos: true
           version: 0.1.6.0
 
   build-and-test:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,15 +14,26 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  generate-matrix:
+    name: "Generate matrix from cabal"
+    outputs: 
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Extract the tested GHC versions
+        id: set-matrix
+        uses: kleidukos/get-tested@v0.1.6.0
+        with:
+          cabal-file: dotenv.cabal
+          ubuntu: true
+          version: 0.1.6.0
+
   build-and-test:
+    name: GHC ${{ matrix.ghc }} on ${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


DavidMazarro:
In addition to the GHC versions found in the `tested-with`, since we also had a manual GHC 9.2 version with a `macos-latest` runner, I tried including that too. However, from the research I've done it looks like GitHub Actions doesn't support the addition of runners to dynamically generated matrices, so I couldn't do it with by using `includes`. I couldn't find an easy way to extend the generated JSON either, I found some examples online but they envolved a mix of tricks with different commands and `jq`. If anyone knows how we could add the GHC 9.2 + macOS runner to the matrix generated by the `get-tested` action, let me know.

That said, should consider if we want to add the `macos-latest` runner to the matrix generation, which means it would run on all GHC versions. I believe macOS runners are also free for open source repositories. So unless the pricing is different for open source repos under organizations, it shouldn't cost us any extra to spin up a `macos-latest` runner for all of the GHC versions instead of just one, and we wouldn't have to keep track and update that isolated version in our CI workflow. Let me know your thoughts on this 🙂

CristhianMotoche:
Hi @DavidMazarro We already have 5 ubuntu instances running. If we add macos to the matrix that will be 10. It seems a little bit too much. Let's look for an easy way to run macos for the last version of GHC. If we don't find one then we may want to keep the previous set up.

DavidMazarro:
After discussing internally, we have decided that we will keep parity between Ubuntu and macOS runners, and we will be cutting the amount of GHC versions we test with. I updated the PR description with more details.

PR #191, opened 2024-03-08
Updating CI to new `get-tested` version (v0.1.7.0)

DavidMazarro:
I opened this issue https://github.com/Kleidukos/get-tested/issues/32 in the `get-tested` repo with a feature request to support setting the specific version of the OS runner we want to run; prior to the v0.1.7.0 version, the runners were all set to `*-latest`. Thanks to @Kleidukos' work, this is now supported in the most recent v0.1.7.0 version of `get-tested`. This PR changes our workflow to this new version of the action.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ jobs:
         with:
           cabal-file: dotenv.cabal
           ubuntu-version: latest
-          macos-version: latest
+          macos-version: 14
           version: 0.1.6.0
 
   build-and-test:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,8 +25,8 @@ jobs:
         uses: kleidukos/get-tested@v0.1.7.0
         with:
           cabal-file: dotenv.cabal
-          ubuntu: true
-          macos: true
+          ubuntu-version: latest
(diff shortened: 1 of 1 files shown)


DavidMazarro:
I've set the `macos-version` to 14 for now with two purposes:
1. Testing that the new version of the action works as expected
2. Out of curiosity, testing the Apple Silicon runner (and see if the library builds successfully)
After that, I'll revert it to `macos-latest`. It would be interesting to ensure compatibility with Apple Silicon, but the runner is still in beta, so I don't think it's advisable using it for now.

DavidMazarro:
Probably not a big surprise, but the Apple Silicon macOS runner doesn't play nicely with GHC 8; it complains about the version of LLVM. When run just with GHC 9.6 and 9.8, it works fine and the library builds without issues.

After testing that, I have reverted back to the `latest` runner.

DavidMazarro:
I think the CI is currently failing because some dependency cache for the Apple Silicon runner was created, and now the `macos-latest` runner is having issues building dependencies with that dependency cache. I'll take a look into it.

CristhianMotoche (approved):

CristhianMotoche:
@all-contributors please add @DavidMazarro for code

Issue #186, opened 2023-11-28
parseFile doesn't seem to escape characters as expected


Martinsos:
Here is an .env file I used to illustrate the potential issue:
```
TEST1="a\nb"
TEST2='a\nb'
TEST3=a\nb
TEST4="a\\nb"
```

Now, when I run `parseFile "testenv"` in `ghci`, I get the following output:
![image](https://github.com/stackbuilders/dotenv-hs/assets/1536647/10690d63-deb6-4674-85a2-d730c14f8685)

Formatted:
```hs
[ ("TEST1","anb"),
  ("TEST2","anb"),
  ("TEST3","anb"),
  ("TEST4","a\\nb"),
]
```

This is not the output I would expect though.

Output I would expect is the following:
```hs
[ ("TEST1","a\nb"),
  ("TEST2","a\nb"),
  ("TEST3","a\nb"),
  ("TEST4","a\\nb")  -- OK I am not 100% sure what to expect here, this might be ok.
]
```

Is my expectation wrong here, or is this potentially a bug in the Dotenv?

I am using `dotenv ^>= 0.10.0`.

Thanks!

Martinsos:
Thanks @CristhianMotoche !

I don't care much about the "\\n" case, but "\n" being parsed as "n" is quite tricky, it makes it impossible to enter a string with a newline inside the env (and for example, Google gives you PEM key as a secret and it has newlines!). The only solution we found so far was to encode it with base64, putting it in .env like that, and then decoding it after being red for dotenv.

Let me know if I can help in some way!

CristhianMotoche:
Hey @Martinsos I was able to reproduce the issue. Indeed, it seems something wrong in the parser. FWIW, I think you can add new lines instead of the `\n` character to represent new lines. For example:

Given this `.env`:

```
BAR="a
b"
```

`parseFile` will return:

```hs
ghci> parseFile ".env"
[("BAR","a\nb")]
```

I hope that works as a workaround until we find a solution for this issue.

PR #192, opened 2024-06-11
fix: Correctly parse newline and multiline strings in dotenv files

FranzGB:
# Proposed Changes

- **test: Introduced tests to validate correct handling of newline characters `\n` and multiline strings in dotenv files.**
- **fix: Implemented parsing of newline and multiline strings in dotenv files accurately**

This pull request addresses issue #186. I wasn't sure if we should parse it as a newline or as a literal string. I decided to go with the latter. But happy to change it if needed. Let me know if you have any feedback.

Additionally, I noticed that our current tests depend heavily on the index of parsed (key, value) pairs, which could make them fragile and prone to breaking with minor changes in the fixture. I propose that we refactor these tests to make them more robust and resilient to changes. I look forward to hearing your thoughts on this.

### DIFF
--- a/src/Configuration/Dotenv/Parse.hs
+++ b/src/Configuration/Dotenv/Parse.hs
@@ -25,7 +25,7 @@ import           Text.Megaparsec                     (Parsec, anySingle,
                                                       between, eof, noneOf,
                                                       oneOf, sepEndBy, (<?>))
 import           Text.Megaparsec.Char                (char, digitChar, eol,
-                                                      letterChar, spaceChar)
+                                                      letterChar, spaceChar, string)
 import qualified Text.Megaparsec.Char.Lexer          as L
 
 type Parser = Parsec Void String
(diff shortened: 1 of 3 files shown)


src/Configuration/Dotenv/Parse.hs
```diff
@@ -85,8 +85,9 @@ interpolatedValueCommandInterpolation = do
       symbol = L.symbol sc
 
 literalValueFragment :: String -> Parser VarFragment
-literalValueFragment charsToEscape = VarLiteral <$> some (escapedChar <|> normalChar)
+literalValueFragment charsToEscape = VarLiteral <$> some (newlineChar <|> escapedChar <|> normalChar)
   where
+    newlineChar  = string "\\n" >> return '\n'
```
CristhianMotoche, line 90: @FranzGB I see `normalChar` parses any char that doesn't need escape. Do you know why that doesn't include the `\n`?

CristhianMotoche (commented):
@FranzGB LGTM! Although, I would like to understand a little bit more why this wasn't parsing the `\n` before.

src/Configuration/Dotenv/Parse.hs
FranzGB: Hi Cris! Yes, basically this line `escapedChar = (char '\\' *> anySingle) <?> "escaped character"` is eating all the `\` not treating `\n` as a single character and leaving just the n as that is a normalChar.

FranzGB: I do have another idea on how could we achieve this without adding a new function here `(newlineChar <|> escapedChar <|> normalChar)` but it will have to require modifying `escapedChar` function. Maybe we can catch up on this WDYT?

CristhianMotoche (approved):
LGTM!

Issue #193, opened 2024-07-06
Fix build workflow for macOS


fm7-1:
The following [build](https://github.com/stackbuilders/dotenv-hs/actions/runs/9458974662/job/26055348810) is failing for macOS. This seems to be caused by the change in the macOS version for the `macos-latest` runner, which is now based on ARM.

PR #194, opened 2024-07-06
fix: Set the macOS runner in the CI to use macOS 13

fm7-1:
This PR solves issue #193 by setting the macOS version to `13`, which is still based on x86.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,7 +3,7 @@ name: Build
 on:
   push:
     branches:
-      - main
+      - fm_fix_ci
   pull_request:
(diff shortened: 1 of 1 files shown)


fm7-1:
Hello @CristhianMotoche, I've opened this PR to fix the error with the build. Some additional effort will be required to make it work with macOS 14, so I think the solution for now would be to use version 13. Would you mind taking a look, please?

CristhianMotoche (approved):
LGTM!

CristhianMotoche:
@all-contributors add @fm7-1 for code

Issue #160, opened 2023-03-14
Introduce a .dotenv.config file to store and read flags for Dotenv CLI


FranzGB:
The current implementation of the Dotenv Command Line Interface (CLI) requires users to provide multiple flags in order to execute a program with specific environment variables. This can result in a lengthy and unwieldy command.

To address this issue, it is proposed that a `.dotenv.config` file be introduced. This file will contain the necessary flags to be passed to the CLI and will be read by Dotenv, allowing users to execute programs with the required environment variables in a more concise and streamlined manner.

This approach is inspired by the behavior of RSpec, a popular testing framework for Ruby. By introducing a `.dotenv.config` file, users will be able to more easily manage the necessary flags for their programs and streamline their workflow.

Example:
```
# .dotenv.config
-x .env.example
-x .env.example.db
-f .env.db 
```

CristhianMotoche:
Hey @FranzGB I've assigned @flandrade to solve this issue. Could you please follow up with her on the development of this feature?

aloussase:
Hi, what is the difference between the `-x` and `-f` flags?

CristhianMotoche:
Hello @aloussase The difference is:

- `-x`(short for `--example`) - Example file. Those files have environment variables (not values) and they're used to enforce that the variables defined within them exist before continuing with the program execution
- `-f`(short for `--dotenv`) - Dotenv file. It's the file that will contain the environment variables (names and values) that will be loaded before a program execution.

I hope that is clear. I think we'll need to revisit our README.md to make this clearer.

PR #197, opened 2024-10-05
Parse configuration file from current directory for additional flags

aloussase:
This is tackling #160 

I added a separate record `Flags` to hold the flags that can be parsed from a configuration file, tentatively `dotenv.config` in the current directory.

At first, I tried to reuse the `Options` record, but this became messy when dealing with the `program` argument. I could not find a non-hacky way to deal with that and thought that it would be best to have a separate parser for the flags.

Having said that, I extracted the option parsers to separate definitions to reuse as much code as possible and also so that the parsers for flags and options keep in sync.

`applyFlags` takes care of combining the flags from command line arguments and the configuration file. `<>` for list arguments is a natural combination, and I believe `||` for booleans also works correctly.

Some execution examples:

```
# dotenv.config
-x .env.example
-f .env
```

```
# .env.example
API_URL=
```

```
# .env
```

Command:

```
dotenv echo hello
```

Output: 

```
dotenv: The following variables are present in .env.example, but not set in the current environment, or .env: API_URL
CallStack (from HasCallStack):
  error, called at src/Configuration/Dotenv.hs:82:16 in dotenv-0.12.0.0-inplace:Configuration.Dotenv
```

Another example:

```
# dotenv.config
echo hello
```

Command:

```
dotenv echo hello
```

Output:

```
There were errors while parsing the configuration file
Invalid argument `echo'

Usage: dotenv [-v|--version] [-f|--dotenv DOTENV] [-x|--example DOTENV_EXAMPLE] 
              [-o|--overload] [--verbose] [--dry-run] [-D|--no-dups]

  Runs PROGRAM after loading options from FILE
```

I print a special message when errors come from the config file so the user knows where to look.

I might be missing something, what do you think?

Also, I had to add `directory` as a dependency to check for the file existence, but if there is a better way to do this without adding an extra dependency that would be great.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -14,7 +14,6 @@ import           Paths_dotenv         (version)
 import           Configuration.Dotenv (Config (..), defaultConfig, loadFile)
 import           System.Directory     (doesFileExist)
 import           System.Exit          (exitWith)
-import           System.IO            (readFile')
 import           System.Process       (system)
 
 data Options = Options
@@ -158,7 +157,7 @@ loadFlagsFromConfig = do
     configExists <- doesFileExist configFile
     if not configExists then return defaultFlags
     else do
-        arguments <- words <$> readFile' configFile
+        arguments <- words <$> readFile configFile
         case execParserPure defaultPrefs (mkOpts flagsP) arguments of
             result@(Failure _) -> do
                 putStrLn "There were errors while parsing the configuration file"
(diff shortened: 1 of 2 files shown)


CristhianMotoche (commented):
@aloussase Thanks a lot for your contribution! It looks great. Thank you so much for putting effort on this. I really like it! I was wondering if you can take a look at the issue in the CI since that's stopping us from merging this PR.

aloussase:
Hi @CristhianMotoche , fixed the CI issue. Let me know if you need anything else :)

CristhianMotoche (approved):
LGTM! Thank you @aloussase

CristhianMotoche:
@all-contributors please add @aloussase for code

PR #201, opened 2024-12-10
chore: Update env variable for semantic-release

CristhianMotoche:


CristhianMotoche:
We're not using semantic release anymore

PR #206, opened 2026-07-22
Remove semantic release

Djojavi:
Remove semantic release in:
 - Deleting release workflow
 - Deleting `.releaserc` 

Actualizing README documentation


CristhianMotoche (changes requested):
@Djojavi Looks good but we should update the [.cabal file](https://github.com/stackbuilders/dotenv-hs/blob/main/dotenv.cabal#L2), too. We don't need to set the 0.0.0.0 as the version. Let's update with the last value of the package in [hackage](https://hackage.haskell.org/package/dotenv): 0.12.0.0

CristhianMotoche (approved):
LGTM!

CristhianMotoche:
@all-contributors please add @Djojavi for code

CristhianMotoche:
@all-contributors please add @Djojavi for code

